PR #31, opened 2020-01-01
Add support for running oasis in a docker container

trashhalo:
Allow running oasis inside docker.

* Adds a dockerfile that starts oasis and ssb
* Moves code that starts the ssb server to its own file

s6-overlay is a small init system that lets you start and manage multiple services inside one docker container. https://github.com/just-containers/s6-overlay


trashhalo:
One thing I haven't quite figured out is how to run accept a pub invite inside the container. Id like to add that to the readme.

Its going to look something like this (after starting the container):
```shell
docker exec -it name_of_container command-to-accept-pub scuttle.us:8008:@ihnyNrsm+7pqHkhnF4u5CkU5VSGlK6YNqRDfHwiUPWM=.ed25519~qdS6CdOhqE1oyNL6d8VfGDrD9/VJDPoKQ7twlcQNaHc=
```

christianbundy:
Wow, this looks great! I'd love to merge this. My only curiosity is about this change:

> Moves code that starts the ssb server to its own file

Is this a workaround for something Docker-related, a refactor for clarity, or something else?

Maybe useful context: I've been trying out a specific shape for the dependency graph ([docs](https://github.com/fraction/oasis/blob/master/docs/contributing.md#always-forward-never-back)) but realized that I've accidentally put some of the heaviest modules in the leaf nodes:

- `/src/pages/models/lib/cooler.js`
- `/src/pages/views/components/template.js`

It seems like the code that gets used the most often should probably be toward the beginning of the dependency graph instead of the end, and I agree that something like `server.js` should be toward the start, but I haven't taken care of it yet.

christianbundy:
Oh, I see. It looks like you may have moved the SSB service into its own file so that you can restart it independently? Maybe you've been seeing those "stream is closed" errors too? I've been thinking that it would be wise to add a retry to the client, and if a retry ever fails then it should automatically restart the server (if that's what it's connected to).

trashhalo:
> Oh, I see. It looks like you may have moved the SSB service into its own file so that you can restart it independently? Maybe you've been seeing those "stream is closed" errors too? I've been thinking that it would be wise to add a retry to the client, and if a retry ever fails then it should automatically restart the server (if that's what it's connected to).

Ya was bumping into stream is closed errors. It seems to not happen when it uses an external ssb server vs booting its own.

In general I think it's cleaner to not have the js manage the lifecycle of a separate ssb server. Seems like a muddying of concerns between Oasis and ssb. I think the purest setup is actually a docker compose with one starting ssb and one starting Oasis and linking the two. But I wanted to created a docker file for people to get up and running quickly. And I reason that one deployable is easier to manage than two.

trashhalo:
> It seems like the code that gets used the most often should probably be toward the beginning of the dependency graph instead of the end

I agree. I was confused by the code in cooler being so deep but being so important to the app.

If you would prefer to keep the ssb management inside node I'd recommend using [pm2](https://pm2.keymetrics.io/) to manage its lifecycle.

Open to any changes to pr! Let me know.

christianbundy:
> I think the purest setup is actually a docker compose with one starting ssb and one starting Oasis and linking the two.

Yeah, that was my original intent behind making the server (Flotilla) a different module with a different name but it seems like complete separation would make it less convenient for most people. Ideally if [this is merged](https://github.com/ssbc/ssb-db/pull/275) we just need the server to run SSB-DB and we can load the rest of our plugins in Oasis. 

> I agree. I was confused by the code in cooler being so deep but being so important to the app.

Thank you, if there's anything else that's super confusing please let me know. It's really great feedback to hear about pain points, and I know there are more than just the ones I know about. :~) Also thanks for the pm2 recommendation.

src/pages/models/lib/cooler.js
```diff
@@ -82,9 +78,7 @@ const handle = new Promise((resolve) => {
   }).catch(() => {
     debug('Initial connection attempt failed')
     debug('Starting Scuttlebutt server')
-
-    server(ssbConfig)
-
+    require('../../../../server')
```
christianbundy: This is the only thing I'm unsure about. I think that the hierarchical `require()` pattern makes it easier to understand what's going on, but [obviously] the server shouldn't be nested this deeply. Is it best to conform to the dependency graph pattern for consistency (and add an incentive to refactor), or is it better to make an exception here? I'm learning toward the former, but I'm down to trust your opinion on this.

Either way, this is on the top of my list for a refactor.

trashhalo: How about I just move server.js into the same folder as cooler and run it from there? Then its all in the same place for you to refactor later

christianbundy: Works great for me, thanks for putting up with the mess. :upside_down_face:

trashhalo: No mess. Code base is pretty tame. You've done a great job so far!

PR #32, opened 2020-01-01
Add bottom margin to lift Comment button

mycognosist:

Aesthetic change: Give the `Comment` button some breathing room (it currently sits on the bottom of the window).

I added `margin-bottom: var(--whole);` to the `main` element in `styles.css`.

I'm not sure if there's a more appropriate place for the style. Maybe it would be better to target the `form` element? Please edit as you desire. Thanks!

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -81,6 +81,7 @@ html {
 
 main {
   margin: 0;
+  margin-bottom: var(--whole);
 }
 
 /* https://www.desmos.com/calculator/3elcf5cwhn */


christianbundy:
Thanks for this patch! I've noticed that problem too but got stuck in some analysis paralysis around "but shouldn't this be applied to `button`" and how that would break the same-row button in the theme chooser (which also needs some design attention IMO).

Anyway: YES! Thank you, merging now.

mycognosist:
Design system coherence - who can even? I'm happy to have been able to nudge things along.

PR #35, opened 2020-01-03
Remove unicorn requirement for systemd service

jedahan:
Make it easier to install and use oasis as a service

Remove hardcoded user and paths to run oasis similarly to how a user would.

Test that ExecStart works with shells other than zsh. This also assumes oasis is in your user path.


contrib/oasis.service
```diff
@@ -1,12 +1,16 @@
...
 [Unit]
 Description=Oasis - Friendly Scuttlebutt Web Interface
 Documentation=https://github.com/fraction/oasis
 After=network.target
 
 [Service]
 Type=simple
-User=unicorn
-ExecStart=/usr/bin/node /home/unicorn/.yarn/bin/oasis --host localhost --port 4515
+ExecStart=%s -l -c 'oasis --host localhost --port 4515'
```
christianbundy: It took me a minute to understand this, could you double-check that I'm reading this right?

- `%s`: It's important to use the user's shell because they might have a custom `PATH`.
- `-l`: We need to invoke a login shell to make sure we source `.profile` / `.bash_profile` / `.zprofile` / etc
- `-c`: The `ExecStart` command isn't a shell, so unless we want to hardcode the path we need to use a shell.

contrib/oasis.service
```diff
@@ -1,12 +1,16 @@
+# cp oasis.service ~/.config/systemd/user
```
christianbundy: What do you think about `mkdir -p ~/.config/systemd/user/` above this line? I had `~/.config` but I'm on a fresh Arch install that didn't have `~/.config/systemd`.

contrib/oasis.service
christianbundy: I just tested this locally and noticed that when running with `sysctl --user` it automatically opens my web browser when the service restarts. What do you think about appending `--no-open`?

docs/install.md
```diff
@@ -29,3 +29,12 @@ yarn
...
 ```
+
+## System service
+
+```shell
+cp contrib/oasis.service ~/.config/systemd/user
+systemctl --user daemon-reload
+systemctl --user enable oasis
+systemctl --user start oasis
+```
```
christianbundy: Note to self: if we add `mkdir -p` then it should probably go here too.

contrib/oasis.service
jedahan: `--no-open` is perfect, will add that shortly

contrib/oasis.service
jedahan: will add that as well.

jedahan:
I also removed docs/install.md because its redundant with the README, lemme know if you'd rather have both.

christianbundy:
Beautiful, this is looking really nice.

I was thinking that the readme should contain the installation methods that don't require that you already have the source (package managers, binaries, etc) and that `docs/install.md` should have instructions on how to download the source code and use it.

Maybe we should move the Docker and systemd install methods to `docs/install.sh` and add a link in the readme? I'd love to end up with something like [Patchwork's readme's install section](https://github.com/ssbc/patchwork#installation).

jedahan:
So are there binaries needed besides npm to install this from source? If not, how about adding a from_source section to the main readme until it really gets complex? Either way, I can revert the doc change since its a bit of a different concern.

christianbundy:
Done, just added a more comprehensive "install from source" and made documents for both the systemd and docker installs available at the top. Please feel free to PR any improvements!

Thanks a bunch, this is really snazzy.

Issue #36, opened 2020-01-03
Add systemd postinstall script


jedahan:
**What version of this package are you using?**
`@latest`

**What problem do you want to solve?**
Make it easier to run on boot if people want to.

**What do you think is the correct solution to this problem?**
Create a postinstall script for npm, something like
```
test -f /run/systemd/system || return
test "$XDG_CONFIG_HOME" || export XDG_CONFIG_HOME=$HOME/.config
export SYSTEMD_USER_HOME="$XDG_CONFIG_HOME"/systemd/user
mkdir -p "$SYSTEMD_USER_HOME"
cp contrib/oasis.service "$SYSTEMD_USER_HOME"/
echo "we copied a systemd service to $SYSTEMD_USER_HOME"
echo "if you'd like to enable oasis on user login, run:"
echo "\n\t systemctl --user enable oasis"
```

**Are you willing to submit a pull request to implement this change?**
ya

christianbundy:
This is a really interesting idea, I'd like to think about this a bit more and hear input from others.

Do you know of any similar projects that do this? I'd love to make it easier to run Oasis as a service but I'm hesitant to add unexpected files. Would this happen only on an `npm --global` install?

Aside: I really appreciate that your scripts respect `$XDG_CONFIG_HOME` and `$SYSTEMD_USER_HOME`, thanks for your attention to detail.

jedahan:
Yeah, this would only run on postinstall globally.

```json
"scripts": {
  "postinstall": "test \"$npm_config_global\" && ./contrib/install-systemd-service.bash || true"
}
```

christianbundy:
Would that work on Windows machines that don't have Bash? I think I'd be comfortable with this as long as it's super robust. Being able to run an SSB service in the background has been a game-changer for me, it feels really good.

jedahan:
We could make it a node script to make it portable

christianbundy:
@RangerMauve

I noticed that you recently published hyperswarm-web, and [the readme](https://github.com/RangerMauve/hyperswarm-web#running-as-a-linux-service-with-systemd) has systemd instructions. Do you think it would be appropriate for that to be done automatically if someone installed with:

```sh
npm --global install hyperswarm-web
```

Also, you might get a kick out of the config that @jedahan put together in #35.

RangerMauve:
@christianbundy Thanks for the ping! I was contemplating doing that, but I got a little spooked by getting it to work cross platform.

[...]

PR #37, opened 2020-01-04
Add install-systemd-service.js

christianbundy:
**What is the purpose of this pull request?** I'm interested in experimenting with #36 and the first step is porting the existing script to JS so that we can run it on systems without Bash. JavaScript is portable.

**What changes did you make? (brief overview)** I replaced the Bash file with a JavaScript file with only one change: the service is automatically enabled by default. I think that if someone is running this command it's safe to bet that it's because they want it enabled.

**Is there anything you'd like reviewers to focus on?** Did I miss anything obvious? I'd love to get a second pair of eyes on this if you'd like to look at it @jedahan. If you don't have time or interest, I'll plan on merging in a few days -- no pressure. Also worth noting, it's getting late-ish here and I'm probably prone to making silly errors. I should probably double-check this in the morning.


christianbundy:
Great idea, I didn't think about that. I'll push another commit and merge once I double-check tomorrow.

PR #39, opened 2020-01-05
Change systemd unit to use `default.target`

christianbundy:

I was having a problem where I'd start my computer but the script wasn't
restarting automatically, and after some searching I found this: https://github.com/systemd/systemd/issues/2690#issuecomment-186973730

It looks like we should be using `default.target` instead of
`multi-user.target`.

Change systemd unit to use `default.target`

Does this look right to you?


jedahan:
Good investigating. Looks good to me.

PR #43, opened 2020-01-07
Add CLI flag for offline mode (using temporary hack with globals)

cinnamon-bun:
This is a new offline branch (the old one `offline-mode` was pretty old, so I started over)

Add `--offline` CLI flag which disables interaction with the SSB network by turning off the conn scheduler.  I believe, but am not 100% certain, this should cover all forms of peer discovery and pub communication.

* Add `--offline` CLI flag
* Get that config info into `server.js` using global variables 😦  since there isn't a way to pass the config down the dependency chain easily.  I tried to do the simplest thing that would work, expecting this to need cleanup later.
* In `server.js`, set `conn.autostart` to `false` when in offline mode
* Small improvements to CLI help text

Going forward, we need a better way to get options down into `server.js`.  In the meantime is this hack acceptable?

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,10 +8,17 @@ const path = require('path')
 
 const config = yargs
   .env('OASIS')
+  .help('h')
+  .alias('h', 'help')
   .usage('Usage: $0 [options]')
   .options('open', {
-    describe: 'Automatically open app in web browser',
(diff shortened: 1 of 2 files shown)


christianbundy:
This is great, thanks so much for the patch! I think this is good to go, and the comments add more documentation that the current architecture is too inflexible for our needs.

PR #46, opened 2020-01-09
Use mediator pattern to simplify architecture

christianbundy:
**What's the problem you solved?** It's hard to pass config options from the CLI to `cooler.js` because they're so far apart. In `master` we use environment variables as globals, but the fact that we have to do this is an architectural code smell.

graph

**What solution are you recommending?** This branch uses an approach similar to the [mediator pattern](https://en.wikipedia.org/wiki/Mediator_pattern) to simplify the application and reduce the number of `require()` hops between the mediator and `cooler.js`.

graph


christianbundy:
One quick note: this breaks the Dockerfile that was depending on `server.js`. If it's important to have a "lite-mode" (no server) or a "daemon-mode" (no HTTP interface) I think we should probably make CLI flags so that we only have to maintain one entry-point into Oasis.

Maintaining one entry-point is a big enough job, and now that we have a simpler architecture I think we should double-down on `index.js` as the entry-point into the application.

christianbundy:
Output from [Tokei](https://github.com/XAMPPRocky/tokei):

```
-------------------------------------------------------------------------------
 JavaScript          Files        Lines         Code     Comments       Blanks
-------------------------------------------------------------------------------
 ./http.js                           60           40            7           13
 ./cli.js                            37           35            0            2
 ./ssb.js                           145          119           14           12
 ./index.js                         519          438           16           65
 ./views/template.js                 58           52            0            6
 ./views/index.js                   299          269            1           29
 ./views/post.js                    164          129           11           24
 ./models/markdown.js                38           31            1            6
 ./models/index.js                 1024          849           28          147
-------------------------------------------------------------------------------
 Total                   9         2344         1962           78          304
-------------------------------------------------------------------------------
```

I'm tempted to refactor the models to be more like [Choo stores](https://choo.io/docs/stores), but I'd prefer to avoid making this PR any bigger.

christianbundy:
Here's a small tour of how `--offline` is threaded through the module:

https://github.com/fraction/oasis/blob/e891a841afbd61e480acc4a688edf63b15a5d6a8/src/index.js#L8

https://github.com/fraction/oasis/blob/e891a841afbd61e480acc4a688edf63b15a5d6a8/src/index.js#L41

https://github.com/fraction/oasis/blob/e891a841afbd61e480acc4a688edf63b15a5d6a8/src/ssb.js#L91-L106

Two files, one require (index -> ssb), and zero environment variables.

christianbundy:
Thanks for the review! I'll merge now, hopefully this makes it easier to get oriented with the code.

PR #50, opened 2020-01-11
Fix Multiserver behavior and remove SSB-Config workaround

christianbundy:
**What's the problem you solved?** The Multiserver module has a bug where if you listen on a meta-address like `0.0.0.0` or `::` and ask which IP addresses it's listening on, it only reports one. I wrote a fix for this in SSB-Config, which translates those meta-addresses into a bunch of individual addresses, but it's a duct-tape solution that I'd like to move past. I've opened pull requests to fix the Multiserver behavior and remove the SSB-Config workaround, but the Multiserver change has been in review since May.

**What solution are you recommending?** I think we should work on [improving both the contribution and review process](https://oasis-demo.fraction.io/thread/%25N4G%2BiskozCSnLYgzMCPozu7xuU2Sh%2BcM3gzGbcgsU%2Bk%3D.sha256#%25N4G%2BiskozCSnLYgzMCPozu7xuU2Sh%2BcM3gzGbcgsU%2Bk%3D.sha256), but until that's solved @staltz [suggested](https://github.com/ssbc/multiserver/pull/49#issuecomment-573116776) that we do some testing in production to make the review process easier.


christianbundy:
Seems like these aren't going to get merged, please see the Multiserver PR for more context.

PR #52, opened 2020-01-11
Add follow and unfollow

christianbundy:
**What's the problem you solved?** It's frustrating to have to open the CLI or Patchwork just to follow / unfollow feeds.

**What solution are you recommending?** Oasis should handle this natively!

### DIFF
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -113,6 +113,29 @@ module.exports = (cooler) => {
   }
 
   models.friend = {
+    // TODO: Refactor `follow` and `unfollow` to share code.
+    follow: async (feedId) => {
+      const ssb = await cooler.connect()
+
+      const content = {
+        type: 'contact',
+        contact: feedId,
+        following: true
+      }
+
+      return cooler.get(ssb.publish, content)
+    },
+    unfollow: async (feedId) => {
+      const ssb = await cooler.connect()
+
+      const content = {
+        type: 'contact',
+        contact: feedId,
+        following: false
+      }
+
+      return cooler.get(ssb.publish, content)
+    },
     isFollowing: async (feedId) => {
       const ssb = await cooler.connect()
       const { id } = ssb
@@ -160,6 +183,7 @@ module.exports = (cooler) => {
         return 'you are following and blocking (!)'
       }
     }
+
   }
 
   models.meta = {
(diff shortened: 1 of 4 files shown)


jedahan:
I made follow and unfollow idempotent, not sure if there are subtle implications, but this was so nice to have.

christianbundy:
LGTM! I've merged `master` into this branch, if you approve it then I'd be happy to merge (or you can approve + merge, makes no difference to me).

src/models/index.js
```diff
@@ -126,6 +126,29 @@ module.exports = (cooler) => {
...
+        contact: feedId,
+        following
+      }
+
+      return cooler.get(ssb.publish, content)
+    },
+    follow: async (feedId) => {
+      const isFollowing = await models.friend.isFollowing(feedId)
+      if (!isFollowing) {
+        await models.friend.setFollowing({ feedId, following: true })
```
jedahan: It would be nice if I could use `this` here, to kinda make it clearer. But I'm not sure how that works in an object literal construction.

jedahan:
I just have one nitpick that I'm not sure how to work around, but if you are cool with it I'll add the feature to the changelog and then merge.

src/models/index.js
christianbundy: I could probably be persuaded, but lately I've been purposefully avoiding `this` because it reduces the amount of magic JavaScript behavior that beginners have to learn. If you'd prefer, we could turn `setFollowing()` into its own little utility function, which isn't exported to `models.friend`, but I'm quite happy with the way that you wrote it.

jedahan: yeah if you feel good about it, i'll add to the changelog and merge

christianbundy: Yep, LGTM!

jedahan:
Interesting that the travis ci builds clone the repo with depth 50, maybe it should just be 1?

christianbundy:
Yeah, 1 seems fine, but [the docs](https://docs.travis-ci.com/user/customizing-the-build/#git-clone-depth) say:

> Please note that if you use a depth of 1 and have a queue of jobs, Travis CI won’t build commits that are in the queue when you push a new commit.

I think that's fine? Not sure. I'm curious what the speedup between 1 and 50 would be. If it's a problem you'd like to solve: :+1:

PR #53, opened 2020-01-12
filter messages outside time range

trashhalo:
If you stand up a fresh sbot and join a pub you get flooded with messages. Then your /popular/day shows you content from 2 years ago. I believe this to be happening because the `gt:` option is filtering on sync time and not message value time. 

After I made this change /popular/day was showing content from about a week ago in time. I assume this to be because of users interacting with week old content in the current day. I feel fine with this as a outcome.

Adding a message filter that filters on the timestamp inside the message.

### DIFF
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -629,7 +629,8 @@ module.exports = (cooler) => {
             return typeof msg.value.content === 'object' &&
         typeof msg.value.content.vote === 'object' &&
         typeof msg.value.content.vote.link === 'string' &&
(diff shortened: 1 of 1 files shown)


christianbundy:
Cool, thanks for the patch!

In my pubs I've resolved this problem via `fix-flumelog-offset`, specifically with my branch that fixes all of the timestamps, but I'm happy to merge this patch until we have a better solution. My only major concern is that if you've just synced a bunch of messages then this will make the 'popular' pages much slower, but 'slow and correct' is probably better than 'fast and incorrect'.

Issue #54, opened 2020-01-14
Let's have a development process


christianbundy:
**What's the problem you want to solved?** Learning a development process is hard work and it's easier when we can outline our needs more explicitly. It would be great to have a development process in this app (that I'd love to also add to SSB-CLI and friends).

**Is there a solution you'd like to recommend?** Since reading about the [Collective Code Construction Contract](https://github.com/zeromq/rfc/tree/master/42) I've been really itching to try it out. There's more info about it [here](https://hintjens.gitbooks.io/social-architecture/content/chapter4.html), but the gist is "patches should solve problems and pass tests, maintainers should merge all correct patches ASAP after review for **correctness only**". I don't want to be a source of truth for this module and it's important to me that I'm not a bottleneck.

The only changes to C4 that I'd propose: 

> A patch commit message MUST consist of a single short (less than 50 characters) line stating the problem ("Problem: ...") being solved, followed by a blank line and then the proposed solution ("Solution: ...").

Commits in this format would be rad, but I can't imagine telling someone to amend their commit to follow this format. I think I'd like for "MUST" to be replaced with "SHOULD".

> The project SHALL NOT use topic branches for any reason. Personal forks MAY use topic branches.

In my experience the project repository is a very convenient place to push your topic branches. I'd be comfortable removing this rule.

> Maintainers SHOULD NOT merge their own patches except in exceptional cases, such as non-responsiveness from other Maintainers for an extended period (more than 1-2 days).

I don't know about this one, to be honest. It's probably fine if you have a bunch of hyper-active maintainers, but right now I think this would mean that I'd be merging lots of my own PRs.

cc: @jedahan @trashhalo @cinnamon-bun @stripedpajamas @mycognosist @aadilayub

jedahan:
Thanks for starting this!

I might even go a step further, and allow for merging of branches that are 'incorrect' but do not break anything. In other words, do not solve the problem, but are a starting point to solving a problem.

I like the change for MUST, I am not sure about the change for branches. Personally it feels good but I'm gonna print out the book and read over it before forming an opinion.

christianbundy:
> In other words, do not solve the problem, but are a starting point to solving a problem.

I think I'm with you, but could you give an example to make sure I'm on the same page with you? As long as it passes tests and it's an improvement (i.e. it solves *some* problem), I think that's a solid candidate for merging.

> Personally it feels good but I'm gonna print out the book and read over it before forming an opinion.

[...]

PR #62, opened 2020-01-19
Add note about supported Node.js versions

christianbundy:
**What's the problem you solved?** I've had a few friends have trouble installing Oasis with an older version of Node.js.

**What solution are you recommending?** Add supported versions to the readme in a way that we don't have to update (like version numbers) and link to the Node.js website for people who want to learn more about the release schedule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Most people should install Oasis with [npm](https://npmjs.org/).
 npm --global install @fraction/oasis@latest
 ```
 
+Please make sure that your Node.js version is the [**current** or **active LTS** release](https://nodejs.org/en/about/releases/).
+
 Want more? Check out [`install.md`](https://github.com/fraction/oasis/blob/master/docs/install.md).
 
 ## Resources


cinnamon-bun:
After approval from code reviewers, is it the responsibility of the original author to hit the Merge button?

christianbundy:
I'd be happy to do it either way. I think my gut feeling is that the person contributing the patch shouldn't be the person merging it, but GitHub doesn't give us an easy way to say "the original author shouldn't merge their own patches". I enabled approvals to make it harder to merge your own patches, but it also feels silly to make people approve and merge. Maybe we should remove the 'require approvals' thing?

jedahan:
this might work: https://github.com/github/platform-samples/blob/master/pre-receive-hooks/block_self_merge_prs.sh

jedahan:
might only be for enterprise :(

christianbundy:
Yeah, unfortunately only for enterprise, but if we moved off of GitHub...

jedahan:
I would love to move off of github...
... BUT if we aspire to the same goal as C4 (that is, # of contributors) I think it would be tricky.

Though I've always been interested in trying glitch as the main center of a community

christianbundy:
Maybe worth discussing more in #54? FWIW I think we should continue to be accessible on GH because it's Where The People Are, but I'm interested in making sure it isn't a single source of truth for the project.

christianbundy:
Thanks!

cinnamon-bun:
Merged!

How about "anyone can hit the Merge button if the conversation has been quiet for a couple days or we think the relevant people have all seen it"?

Otherwise the first reviewer could merge it before others have a chance to review it.

The person most likely to come back and check on it 2 days later is the original author, since they're excited to get it merged, but it could be anyone.

christianbundy:
> I feel it doesn't matter who actually clicks Merge once code review has happened.

I could do that. The one thing I do like about 'the author doesn't merge' is that it's reflected in Git -- I'd love if the merge commit was made by the person who reviewed the code, but I could also live without it. :man_shrugging: I like the separation between C4's 'contributor' and 'maintainer' hats, and merging your own patch feels like a layer violation. Maybe it's fine for now though?

> How about "anyone can hit the Merge button if the conversation has been quiet for a couple days or we think the relevant people have all seen it"?

That works for me. My preference would be to merge more often rather than less, so things are merged by default unless they have a big problem, but I'm also probably the person making lots of PRs so I'm biased.

jedahan:
I would really love to see how strict we can be with C4, and discuss the real-world effects as it goes on. Essentially, there is a tiny bit of tension, especially when a project has fewer contributors, to "mean time between merge" and "the maintainer for a patch cannot be the contributor of that patch". I assume MTBM will decrease as number of maintainers increase. I also assume that it feels bad to leave a patch "languishing". For now, we can assume a best-effort as we get more comfortable with different roles.

We can also share somewhere "practical friction implementing C4 on github". Or maybe we create a simple merge-bot, that merges automatically after 72 hours as long as something has been approved? I dunno.

PR #65, opened 2020-01-22
Experiment with common-good module

christianbundy:
**What's the problem you solved?** This module maintains a bunch of linter configurations and opinions. Also, `standard`'s code formatting is sort of niche and only works on JavaScript.

**What solution are you recommending?** I think would be better if these opinions were abstracted into their own module (common-good). Also, I'd like to use Prettier to format code. It's much more popular and also formats CSS / JSON / Markdown / etc), which lets us use ESLint to *only* check for code problems instead of style issues.

For what it's worth, I don't plan to program in the Prettier style, but I'm happy to have my code formatted that way before committing. The big JS style changes are:

- **Semicolons.** This is actually kinda nice because it means that you can make modifications without having to look at the adjacent lines to watch out for ASI (automatic semicolon insertion) hazards.
- **Convenient quotes**. Prettier chooses whichever quote style requires fewer `\"escapes\"`, which is also kinda nice.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,16 @@
+module.exports = {
+  env: {
+    commonjs: true,
+    es6: true,
+    node: true
+  },
+  extends: "eslint:recommended",
+  globals: {
+    Atomics: "readonly",
+    SharedArrayBuffer: "readonly"
+  },
+  parserOptions: {
+    ecmaVersion: 2018
+  },
+  rules: {}
+};
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -59,19 +59,9 @@
 
 html {
   display: flex;
-  font-family:
-    system-ui,
-    -apple-system,
-    BlinkMacSystemFont,
-    "Segoe UI",
-    "Roboto",
-    "Oxygen",
-    "Ubuntu",
-    "Cantarell",
-    "Fira Sans",
-    "Droid Sans",
-    "Helvetica Neue",
-    sans-serif;
+  font-family: system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI",
+    "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans",
+    "Helvetica Neue", sans-serif;
   justify-content: center;
   font-size: 12pt;
   line-height: 1.5;
@@ -85,12 +75,24 @@ main {
 }
 
 /* https://www.desmos.com/calculator/3elcf5cwhn */
-h1 { font-size: 133%; } /*     4 / 3 */
-h2 { font-size: 115%; } /*     8 / 7 */
-h3 { font-size: 105%; } /*   16 / 15 */
-h4 { font-size: 103%; } /*   32 / 31 */
-h5 { font-size: 101%; } /*   64 / 63 */
-h6 { font-size: 100%; } /* 128 / 127 */
+h1 {
+  font-size: 133%;
+} /*     4 / 3 */
+h2 {
+  font-size: 115%;
+} /*     8 / 7 */
+h3 {
+  font-size: 105%;
+} /*   16 / 15 */
+h4 {
+  font-size: 103%;
+} /*   32 / 31 */
+h5 {
+  font-size: 101%;
+} /*   64 / 63 */
+h6 {
+  font-size: 100%;
+} /* 128 / 127 */
 
 h1,
 h2,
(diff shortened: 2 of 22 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,16 @@
...
+  extends: "eslint:recommended",
+  globals: {
+    Atomics: "readonly",
+    SharedArrayBuffer: "readonly"
+  },
+  parserOptions: {
+    ecmaVersion: 2018
+  },
+  rules: {}
+};
```
christianbundy: (These are all the defaults generated by `eslint --init`.)

jedahan:
It took me a minute to reframe the problem so I understood it better:

**Problem** the oasis project has a lot of lines dedicated to styling and formatting, which is not unique to the project, therefore contributes to noise that can make it harder to contribute. (meta-thought: the phrase "harder to contribute" might be meaningless in the same way "simple" can be meaningless. I hope to avoid using it in the future as it just seems loaded).

**Solution** Move all easily contained, meta-code to its own module.

Now, when I read Experiment, I like that it means 'hey, this can be changed, lets just try it out for a bit'. What would be the impetus for continuing to adopt, or reverting this change? i.e. "After 2 months, we will check in with contributors and ask how they feel about common-good?" or "If we see X issues that are not resolved in common-good, we will revert" or "If we see Y new contributors, we will continue".

I think the ', also' kinda implies there should be a second PR but I understand this makes sense to have both changes in here, and appreciate that you summarized the big ones. I personally prefer standard's style (no semicolon, and generally I just use `` everywhere), but its way more important to just have something that is easy to run, and that there is agreement that there's no way I'd dig my heels in about any particular style.

christianbundy:
Thanks for the quick feedback!

> Now, when I read Experiment, I like that it means 'hey, this can be changed, lets just try it out for a bit'. What would be the impetus for continuing to adopt, or reverting this change? 

The lens I'm currently trying on: anyone at any time can revert this change and I'll be happy to merge it. Maybe someone will hate the `common-good` bundle and revert most of this, leaving Prettier instead of StandardJS. Or maybe they'll try something else. :man_shrugging: If things feel too turbulent we should discuss, but otherwise I'm operating under the principles discussed in #54.

> I personally prefer standard's style (no semicolon, and generally I just use `` everywhere), but its way more important to just have something that is easy to run, and that there is agreement that there's no way I'd dig my heels in about any particular style.

Same, and I'm planning to continue writing code in a StandardJS-like style, but I completely agree with you. I'd like to optimize for how easy it is for new contributors to open a random file and start editing. I'd feel awful if they edited under a line like `f()` with `[1, 2, 3].forEach()` and they had a big error because they hit an ASI hazard. I'm impressed that Prettier provides syntax for most of the languages we use, which makes it harder to disagree about.

Aside: I've considered changing one default, [`--arrow-parent=always`](https://prettier.io/docs/en/options.html#arrow-function-parentheses), because it makes it harder to add a second argument (because you have to write parens), but I'm trying to stick with the default instead of injecting my immediate opinions.

src/assets/style.css
```diff
@@ -85,12 +75,24 @@ main {
...
+} /*   16 / 15 */
+h4 {
+  font-size: 103%;
+} /*   32 / 31 */
+h5 {
+  font-size: 101%;
+} /*   64 / 63 */
+h6 {
+  font-size: 100%;
+} /* 128 / 127 */
```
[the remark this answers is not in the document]
jedahan: Maybe we make a block comment at the top - these sizes derived by the function (function from weblink here), which leads to
`4/3 8/7 16/15 32/31 64/63` ....

src/views/post.js
```diff
@@ -144,21 +137,26 @@ module.exports = ({ msg }) => {
...
-        `❤ ${likeCount}`)),
-      a({ href: url.comment }, 'comment'),
-      (isPrivate || isRoot || isFork) ? null : a({ href: url.reply }, 'reply'),
-      a({ href: url.json }, 'json')
-    ))
-
-  return fragment
-}
+      form(
+        { action: url.likeForm, method: "post" },
```
jedahan: this split is a little frustrating

cinnamon-bun:
@christianbundy Does common-good have a way to fix formatting?  I ran `npx standard --write <myfile>` but I'm not sure that's the recommended way.

christianbundy:
Yup! It's just `common-good fix`. In this repo you can also `npm run fix`. Probably something I should add to the docs.

PR #66, opened 2020-01-22
Create svg favicon

jedahan:
When I pin the tab on firefox, the icon shown is generic. I'd love to see the DESERT ISLAND. Left is currently pinned, right is with this patch.

favicon

Make an svg favicon. I'd be interested in seeing if it renders correctly in different browsers.

### DIFF
--- a/src/assets/favicon.svg
+++ b/src/assets/favicon.svg
@@ -1,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 15 15">
+    <title>oasis favicon</title>
+    <text x="0" y="15">🏝️</text>
+</svg>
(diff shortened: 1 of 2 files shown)


jedahan:
There is also a world in which could be embedded directly, but I timeboxed this work and am out of time :)

christianbundy:
YES. This is so cool, I've been really wanting a favicon for this exact reason but really struggled with pixel art. Plus, I really liked having the icon just be an emoji, which looks different on different platform. This is the best of all possible worlds, thanks for the patch!

jedahan:
Nah I was just trying to distribute the load, so feel free to merge

jedahan:
Note that this is not compat with all browsers. https://caniuse.com/#feat=link-icon-svg
So there can be improvments

Issue #68, opened 2020-01-22
Show peer names in Meta & explain what's happening there


cinnamon-bun:
**What's the problem you want to solved?**
The Meta page shows peers I'm connected to, but only by their pubkey.

**Is there a solution you'd like to recommend?**
Instead of pubkey, show their names.

While we're revamping this, it's also a good time to add a bit of help text.  People tend to be confused about peer network connections vs. social following connections.  I've seen them get scared, saying "who's this person looking at me??  How did they find me?"

Maybe something like...

**Peer connections 💻⚡️💻**
*Your computer is syncing data with these other computers.  It will connect to any scuttlebutt pub and peer it can find, even if you have no relationship with them, as it looks for data from your friends.*
* `@foo`
* `@bar` 

(That's adapted from part of the [scuttlebutt.nz help text](https://scuttlebutt.nz/docs/introduction/detailed-start/#what-are-all-these-pubs-listed-in-the-patchwork-sidebar-i-didn-t-join-those) which I wrote)

cinnamon-bun:
That PR adds help text but not peer names.  I'm not sure how to look up the names.

christianbundy:
> I'm not sure how to look up the names.

Happy to support you working on this. You'll probably want to edit `src/index.js` to use [this code](https://github.com/fraction/oasis/blob/2adcede85f9a1155c9b3bdf0e13007f13846cc5d/src/models/index.js#L31-L39):

```js
const feedName = await about.name(feedId)
```

PR #69, opened 2020-01-22
Add help text in Meta about peer connections

cinnamon-bun:
People are often confused about the difference between peer connections and social/following connections

* Added some help text on the Meta page Peers section as mentioned in #68.
* When there are no peers, say "no peers connected" instead of omitting the Peers section
* Moved Peers section before Indexes because I expect people to look at Peers more often than Indexes


christianbundy:
This looks great, thanks for the patch!

PR #72, opened 2020-01-23
Make Travis faster by removing double-build

christianbundy:
**What's the problem you solved?** Travis builds PRs and branches, which means that local branches
get built twice (!).

**What solution are you recommending?**  Travis should only build `master`, and maybe tags. This commit also contains the changes from `npm --init`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,12 +2,15 @@
   "name": "@fraction/oasis",
   "version": "2.11.0",
   "description": "friendly neighborhood scuttlebutt interface",
-  "repository": "git@github.com:fraction/oasis.git",
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/fraction/oasis.git"
+  },
   "license": "AGPL-3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,7 +45,8 @@
     "ssb-msgs": "^5.2.0",
     "ssb-ref": "^2.13.9",
     "ssb-thread-schema": "^1.1.1",
-    "yargs": "^15.0.0"
+    "yargs": "^15.0.0",
+    "sharp": "^0.23.0"
```
christianbundy: I'm conflicted about this change from `npm init`. The `optionalDependencies` property overrides `dependencies`, so this won't cause any change to our behavior, but the docs say:

> Entries in **`optionalDependencies`** will override entries of the same name in **`dependencies`**, so it’s usually best to only put in one place.

Why does `npm init` do this?!

christianbundy: tbh I'm slightly annoyed but also totally fine with the behavior -- it's weird, but I'm fine with npm defaults as long as that's what we're using :man_shrugging:

PR #73, opened 2020-01-23
Show peers by name, by host, or by key

cinnamon-bun:
Peers might not have a known name

Look in this order and use the first one that's defined: name, hostname, key.

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -174,9 +174,14 @@ exports.metaView = ({ status, peers, theme, themeNames }) => {
     return div(label({ for: id }, key), progress({ id, value: val, max }, val));
   });
 
-  const peerList = (peers || []).map(([, data]) =>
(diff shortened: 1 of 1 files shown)


src/views/index.js
```diff
@@ -237,13 +242,13 @@ exports.metaView = ({ status, peers, theme, themeNames }) => {
...
-        ? ul(peerList)
-        : code('no peers connected'),
-      h3('Indexes'),
+      h2("Status"),
+      h3("Peer Connections 💻⚡️💻"),
+      p(
+        "Your computer is syncing data with these other computers. It will connect to any scuttlebutt pub and peer it can find, even if you have no relationship with them, as it looks for data from your friends."
+      ),
+      peerList.length > 0 ? ul(peerList) : code("no peers connected"),
+      h3("Indexes"),
```
cinnamon-bun: This part changed when I ran "npx standard --write" to apply the code style.  It's only a formatting change.

PR #74, opened 2020-01-23
Change favicon dimensions to fit in tab

christianbundy:
**What's the problem you solved?** Some of the icon was being clipped in my browser. I think that this is because the SVG text actuall extends down under the line (like a `g` or `y`).

**What solution are you recommending?** Change SVG viewBox dimensions and SVG size to fit correctly.

Before and after:

Screenshot from 2020-01-23 13-42-47

### DIFF
--- a/src/assets/favicon.svg
+++ b/src/assets/favicon.svg
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 15 15">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
     <title>oasis favicon</title>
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
This looks good to me.  I'm not listed as a reviewer so I'm leaving it for @jedahan.

I'm amused by this trick of using an emoji inside an SVG :)

christianbundy:
Oh, feel free to merge -- I probably should've just CC'd instead of assigning for review.

jedahan:
Somehow gits merging algorithm reverted the change. Fixing in this PR.

jedahan:
(i wonder if pijul would have done it right...)

PR #75, opened 2020-01-23
Fix private reply with `recps` bug

christianbundy:
**What's the problem you solved?** Trying to reply to some messages sent with Patchbay fail because the schema check is throwing an error.

**What solution are you recommending?** When we encounter `recps` like `{ name, link }`, normalize it to just `link` and publish a well-formed message that passes the schema check.

### DIFF
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -991,7 +991,22 @@ module.exports = cooler => {
         // We default to `null` in case something goes wrong and we receive a
         // private message with no `recps` somehow. This adds a layer of security
         // because SSB-DB won't let us publish a message with `{ recps: null }`.
-        message.recps = lodash.get(parent, "value.content.recps", null);
+        message.recps = lodash
+          .get(parent, "value.content.recps", [])
+          .map(recipient => {
+            if (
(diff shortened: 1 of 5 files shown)


christianbundy:
See also: https://github.com/ssbc/ssb-schema-definitions/issues/2

src/models/index.js
```diff
@@ -991,7 +991,22 @@ module.exports = cooler => {
         // We default to `null` in case something goes wrong and we receive a
         // private message with no `recps` somehow. This adds a layer of security
         // because SSB-DB won't let us publish a message with `{ recps: null }`.
```
cinnamon-bun: This comment doesn't match the code now - it defaults to `[]` instead of `null`.  Is it worth checking for an empty array after all this and setting it to `null` in that case?

christianbundy: Thanks, I'll fix that comment! `[]` is fine too, SSB won't publish a message with `recps` unless `recps` is an array with >= 1 recipient which has to be a valid SSB feed ID. I'll add some more security here anyway, just in case.

PR #76, opened 2020-01-23
Update npm dependencies for new common-good

christianbundy:
**What's the problem you solved?** Common-Good had a bug in the Prettier glob where the recursion
wasn't working correctly. This meant that we weren't actually testing or fixing some of the deeper files.

**What solution are you recommending?** Update CG to the latest version and run `npm run fix` to fix the linter stuff.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "devDependencies": {
     "changelog-version": "^1.0.1",
-    "common-good": "^1.1.17",
+    "common-good": "^1.1.18",
     "husky": "^3.0.5",
     "mkdirp": "^0.5.1",
     "nodemon": "^2.0.2",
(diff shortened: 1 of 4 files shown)


christianbundy:
(Also got to fix a Windows bug in cspell)

package.json
```diff
@@ -38,19 +38,19 @@
     "pull-sort": "^1.0.2",
     "pull-stream": "^3.6.12",
     "require-style": "^1.1.0",
-    "ssb-client": "^4.7.7",
+    "sharp": "^0.23.0",
+    "ssb-client": "^4.8.0",
```
cinnamon-bun: Also bumps ssb-client version, I see

christianbundy:
Thank you! Yep, SSB-Client came too. :+1:

PR #77, opened 2020-01-24
Experimentally add turbolinks.js to mitigate slow page loads

cinnamon-bun:
Some pages load slowly.

I expect this to be controversial!  Don't merge this until we've discussed it.

I experimentally added [turbolinks.js](https://github.com/turbolinks/turbolinks).

It progressively enhances a plain-HTML site by caching pages in browser memory, changing it into a single-page app behind the scenes.  When you navigate to a new page it:
1. intercepts the navigation event
2. immediately shows the cached version
3. re-fetches the page in the background, showing a CSS progress bar while that happens
4. eventually swaps out the cached version for the new version

To try this out: navigate to a slow page, then a fast one, then back to the slow one.

It does make things feel subjectively faster since you can see the old version of a page while waiting for the new version to load.

Browsers with javascript disabled will continue to work as normal.

This would be a single-line change to add a <script> tag, except I had to also **enable same-domain scripts in the content security policy**.  We could hide this all behind a command-line flag, or just not do it.


christianbundy:
Interesting, thanks for opening this PR! I'd be happy to experiment with this.

**Background:** My main priority with avoiding front-end JS is just reducing the number of things that Oasis contributors have to worry about. There's *tons* of complexity in the SSB stack, but interfaces are usually the starting point for learning the stack and I really want to empower beginners with simplicity and accessibility. On the other hand, browsers are *not* simple document viewers and we should expect dynamic behavior from browsers (like spellcheck) or their extensions (like vimium). Front-end JS is going to happen whether we like it or not, but I [obviously] think we should be careful about which front-end JS is within the scope of Oasis.

I like this because it makes slow pages faster, which is nice because life is short. Turbolinks doesn't require any fussy configs or any of our brainpower to keep it up and running, which are both great.

I dislike this in this context because it adds new page states (cached page, fresh page) and a new class of bugs (click 'latest' then 'public' then 'week' -- I think turbolinks is having trouble with our redirect from `/` to `/public/popular/day`). I'm also hesitant because it's another thing to potentially learn / configure, which ranges from "why doesn't the blue match my theme color" to "why does the loading bar hang forever" (could this mask an error message?).

From a macro view, I think that if we merge this we'll need to draw a new line in the sand. Maybe "no JS code unless it's a helper library and gives us zero options for configuration"? I'm not sure. On a long enough timeline I think we'll probably want to discuss a lightweight `<textarea>` enhancement that makes Markdown easier, or maybe some other plugin, which makes me wonder whether we should discuss that sort of thing now-ish.

My first impulse is "people should be able to add custom CSS / JS in their settings", which might be useful to some people, but probably alienating to people who aren't comfortable with software.

**TL;DR:** It's really cool to see these patches, I'm happy to try this experiment together, I found a small bug with how it handles redirects, and I think this an important part of a bigger discussion about front-end JS helpers.

christianbundy:
Another note: I think we'll need to be very careful about loosening the CSP. If someone shares an HTML blob with a `<script>` pointing to another blob URL, I think they would be able to run arbitrary code in your browser. I *think* that since we're sharing a domain, then they could send HTTP POST requests to simulate form submissions and they'd basically own Oasis. It would be very important to avoid this. :upside_down_face: 

I think it'd be safer to move blobs to their own subdomain, either `blobs.localhost` or `<blobKey>.localhost`, but I don't know enough about how other operating systems handle localhost subdomains (and I'm slightly hesitant to add the complexity, since we currently have a CSP).

cinnamon-bun:
@christianbundy thanks for the long thoughts on this!

After a day of using it with Turbolinks I'm no longer a fan, so I'll close this PR:
1. I'd like for a page to hold still after I open it.  I don't like the progress bar animation (which can be changed, yes) or the way the page can jump when the new version is loaded
2. Redirect bug
3. Would make adding other JS more complex and I'd rather spend our javascript budget on a better markdown and mention UX.

Re lines in the sand, it would be useful to have a list of the project's priorities and values ("easy for new contributors to understand"), and the constraints that meet those values ("limited frontend javascript").

Re: custom CSS, yes!  That's a relatively easy onramp for people to start getting under the hood of the software

PR #79, opened 2020-01-25
Use ssb-backlinks to improve /inbox page load

christianbundy:

The /inbox page was being rendered super slowly because it was
reading through tons of messages.

There isn't a way to query the database for "private messages
for me", although maybe there should be, but one way we can get
something close is querying for "messages that reference me". Every
message that's encrypted for us will have a `.value.content.recps`
property that includes our feed ID, so we just have to filter out the
public messages and we're about 4 times faster than the previous
implementation.

```console
$ hyperfine 'curl -I http://localhost:4515/inbox' 'curl -I http://localhost:3000/inbox'
Benchmark #1: curl -I http://localhost:4515/inbox
  Time (mean ± σ):      3.352 s ±  0.093 s    [User: 2.0 ms, System: 4.3 ms]
  Range (min … max):    3.231 s …  3.483 s    10 runs

Benchmark #2: curl -I http://localhost:3000/inbox
  Time (mean ± σ):     811.8 ms ±  88.3 ms    [User: 2.7 ms, System: 2.9 ms]
  Range (min … max):   709.1 ms … 972.5 ms    10 runs

Summary
  'curl -I http://localhost:3000/inbox' ran
    4.13 ± 0.46 times faster than 'curl -I http://localhost:4515/inbox'
```

(I run my development server on port 3000 and the latest stable release on port 4515.)

### DIFF
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -1036,22 +1036,22 @@ module.exports = cooler => {
 
       const options = configure(
         {
-          type: "post"
+          query: [{ $filter: { dest: ssb.id } }]
         },
         customOptions
       );
 
-      const source = await cooler.read(ssb.messagesByType, options);
+      const source = await cooler.read(ssb.backlinks.read, options);
 
       const messages = await new Promise((resolve, reject) => {
         pull(
           source,
+          // Make sure we're only getting private messages that are posts.
           pull.filter(
-            (
-              message // avoid private messages (!)
-            ) =>
+            message =>
               typeof message.value.content !== "string" &&
-              lodash.get(message, "value.meta.private")
+              lodash.get(message, "value.meta.private") &&
+              lodash.get(message, "value.content.type") === "post"
           ),
           pull.unique(message => {
             const { root } = message.value.content;


cinnamon-bun:
**(oasis v2.11.0)**
  Time (mean ± σ):      4.917 s ±  0.176 s    [User: 4.2 ms, System: 3.9 ms]

**(just before this commit)**
  Time (mean ± σ):      4.542 s ±  0.137 s    [User: 4.0 ms, System: 3.7 ms]

**(this commit)**
  Time (mean ± σ):     960.6 ms ±  58.6 ms    [User: 3.7 ms, System: 3.7 ms]

awesome work 🐎

christianbundy:
Thanks for merging these patches! Btw, just found this fun hyperfine option: `--export-markdown`. Looks like it makes a fancy little table.

```sh
hyperfine --export-markdown /dev/stdout 'curl localhost:3000/inbox' 'curl localhost:4515/inbox'
```

|:---|---:|---:|---:|---:|
| `curl localhost:3000/inbox` | 856.2 ± 86.0 | 775.2 | 1019.5 | 1.00 |
| `curl localhost:4515/inbox` | 3405.8 ± 196.3 | 3227.7 | 3857.9 | 3.98 ± 0.46 |

Issue #21, opened 2019-12-11
Liking a post throws an error


aadilayub:
**What version of this package are you using?**
2.9.1

**What operating system, Node.js, and npm version?**
Fedora 31, Node v12.13.1, npm 6.12.1

(I also have patchwork running in the background)

**What happened?**
I clicked "like" on a post, and got redirected to a page with this text:

```
Error: non-error thrown: {"message":"no async:tangle,branch","name":"Error","stack":"Error: no async:tangle,branch\n    at Object.localCall (/tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/muxrpc/local-api.js:29:13)\n    at Object.<anonymous> (/tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/muxrpc/local-api.js:37:22)\n    at Object.request (/tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/muxrpc/stream.js:48:17)\n    at PacketStream._onrequest (/tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/packet-stream/index.js:161:17)\n    at PacketStream.write (/tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/packet-stream/index.js:134:41)\n    at /tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/muxrpc/pull-weird.js:56:15\n    at /tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/pull-stream/sinks/drain.js:24:37\n    at /tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/pull-goodbye/node_modules/pull-stream/throughs/filter.js:17:11\n    at Object.cb (/tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/packet-stream-codec/index.js:111:11)\n    at drain (/tmp/.mount_patchwzywQgP/resources/app.asar/node_modules/pull-reader/index.js:39:14)"}
    at Object.onerror (/home/aadil/.npm-global/lib/node_modules/@fraction/oasis/node_modules/koa/lib/context.js:113:40)
    at onerror (/home/aadil/.npm-global/lib/node_modules/@fraction/oasis/node_modules/koa/lib/application.js:163:32)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

**What did you expect to happen?**
The post gets liked, maybe the heart icon animates or turns pink

[...]

PR #83, opened 2020-01-26
Fix tangle bug when Patchwork is used as a server

christianbundy:
Problem: Patchwork is missing a plugin that we need to set the `branch`
property when posting a message. This property is important because it
helps us sort threads, so we're throwing an error when it isn't
available. See: https://github.com/fraction/oasis/issues/21

Solution: HACK THE ~~PLANET~~ API. This commit injects the plugin we
need via Oasis, which is a bit of a duct tape solution but it *is* a
solution.

### DIFF
--- a/src/ssb.js
+++ b/src/ssb.js
@@ -59,12 +59,14 @@ const rawConnect = () =>
         if (err) {
           reject(err);
         } else {
-          // HACK: SSB-Tangle isn't available in Patchwork, but we want that
-          // compatibility. This code automatically injects SSB-Tangle into our
-          // stack so that we don't get weird errors when using Patchwork.
-          //
-          // See: https://github.com/fraction/oasis/issues/21
-          api.tangle = ssbTangle.init(api);
+          if (api.tangle === undefined) {
+            // HACK: SSB-Tangle isn't available in Patchwork, but we want that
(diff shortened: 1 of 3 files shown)


christianbundy:
cc: @aadilayub  @georgeowell, I've tested locally and can confirm that this seems to solve the problem you've both reported. Sorry for dropping the ball on this fix for so long.

src/ssb.js
```diff
@@ -58,6 +59,13 @@ const rawConnect = () =>
         if (err) {
           reject(err);
         } else {
+          // HACK: SSB-Tangle isn't available in Patchwork, but we want that
+          // compatibility. This code automatically injects SSB-Tangle into our
+          // stack so that we don't get weird errors when using Patchwork.
+          //
+          // See: https://github.com/fraction/oasis/issues/21
+          api.tangle = ssbTangle.init(api);
```
jedahan: im not super familiar with how init and everything work, but maybe we should check if `api.tangle` is null before assigning it?

christianbundy: Sure! This implementation overwrites the existing `tangle` if it exists, which I think is fine, but we can also check beforehand so that we aren't modifying the object unless we need to.

christianbundy:
Thank you!!!

Issue #89, opened 2020-01-26
Add UX for enabling and disabling local syncing


georgeowell:
This would be the same as running `oasis --offline` on the command line.

christianbundy:
Resolved in #92!

Issue #63, opened 2020-01-19
Sometimes connects to peers but doesn't get new messages


christianbundy:
**What's the problem you want to solved?** https://github.com/staltz/ssb-conn/issues/11

**Is there a solution you'd like to recommend?** Maybe as a stopgap it would be wise to just add a "reset networking" button to the 'meta' page? I've been running `ssb gossip reconnect` in my terminal but I don't want people to have to use a terminal to workaround Oasis problems.

jedahan:
Reset networking is nice, maybe call it 'force refresh' or 'scan for new connections'. One question is how we illuminate when to press the button. Perhaps if, 5 minutes after starting, if the latest message is over 3 days old, and no new messages are found, point out the refresh button?

cinnamon-bun:
I've seen this happen in Patchwork also.

A button on the Meta page sounds good.

christianbundy:
Closing after workaround from #92

PR #92, opened 2020-01-27
Add buttons to 'meta' page to manage networking

christianbundy:
Problem: Sometimes you want to disconnect from the network but you don't want to have to restart Oasis with the `--offline` flag (https://github.com/fraction/oasis/issues/89). Sometimes networking gets stuck and you need to run the equivalent of `ssb gossip reconnect` (https://github.com/fraction/oasis/issues/63).

Solution: Buttons on the 'meta' page that let you start, stop, or restart SSB-CONN whenever you want. Note that this commit includes an update to SSB-CONN, but this version and the previous version both have a bug where hitting 'stop' twice in a row will throw an error. This commit implements a workaround for the bug, but it's something we'll want to remove later once the underlying bug is fixed.


christianbundy:
cc: @georgeowell

christianbundy:
Obligatory screenshot: 
Screenshot from 2020-01-26 17-00-40

cinnamon-bun:
Nice!

Some peers linger in the list when I click "stop".  Clicking "stop" again clears them out.

This overrides `--offline` mode - would be good to document that in the command line help.

I wish it showed the current on/off state of networking next to the buttons.

src/models/index.js
```diff
@@ -206,6 +206,33 @@ module.exports = cooler => {
...
+    },
+    connStart: async () => {
+      const ssb = await cooler.connect();
+      const result = await cooler.get(ssb.conn.start);
+
+      return result;
+    },
+    connRestart: async () => {
+      await models.meta.connStart();
+      await models.meta.connStop();
```
cinnamon-bun: Are these lines swapped?  I expected restart to be stop, then start.

cinnamon-bun:
I tried changing the Stop button to run `ssb.conn.stop` 4 times in a row.  It didn't fix the lingering peers problem.

I wonder if `conn` is actually still connected to the lingering peers?  I haven't seen them disappear though I didn't wait more than a couple of minutes.

src/models/index.js
christianbundy: Ah! These *are* swapped, good catch.

christianbundy:
> This overrides --offline mode - would be good to document that in the command line help.

Yeah, I really don't know either. I added a note about it, but unfortunately don't know enough about the guts of SSB-CONN. Maybe cc: @staltz

> This overrides `--offline mode` - would be good to document that in the command line help.

Hmm, should we just disable this altogether when we're offline? I'm not sure what the right move is here.

staltz:
@christianbundy I find it strange that you (users) have to manually call `start` and `stop`, specially this:

> Sometimes networking gets stuck and you need to run the equivalent of ssb gossip reconnect (#63).

Another thing you may want is to put `{conn: {autostart: false}}` in the ssb config.

christianbundy:
Thanks @staltz.

> Another thing you may want 

That's the behavior of `oasis --offline`. The problem we're bumping into is that when you start the app normally, with networking, when you hit `stop` and refresh the peer list there are usually still peers. Even if you hit stop a bunch of times, the peers don't go away, you have to wait. I think it takes like 30 seconds maybe?

cinnamon-bun:
@christianbundy I don't know if you want to merge this now and fix it later, or hold off until it's understood better.  If you want it merged now that's ok with me.  I'll delegate the merge-button clicking to you so I don't accidentally merge it before you're ready.

staltz:
@christianbundy Can you open an issue in ssb-conn so we can discuss the `stop` behavior?

christianbundy:
Fixed and ready for review!

jedahan:
The stop networking button seems great, the start/restart is trickier because it just says 'no peers connected' but I don't see any feedback of the status. Will merge and capture that in another bug.

Issue #28, opened 2019-12-16
Improve navigation (sidebar?)


christianbundy:
**What version of this package are you using?** 2.9.7

**What problem do you want to solve?** The top navbar is simple, but not very ergonomic (especially on small screens). I think we should improve it, but I'm not sure how.

**What do you think is the correct solution to this problem?** I think a Patchwork-style sidebar is reasonable on desktop, but I'm not sure what to do on mobile. Here's a very rough sketch that wouldn't take very long to implement:

![Screenshot from 2019-12-16 11-48-02](https://user-images.githubusercontent.com/537700/70937945-1acaba00-1ffa-11ea-95c2-645e643c6517.png)

I'm inspired to make a Shellogram-style sidebar that's more interactive, but I'm in some technical debt that would make that difficult right now. I should probably just repay that debt. My mistake was implementing a tree-like dependency graph where the most often-used tools (database connection, main template, etc) are leaf nodes, when *really* I think the leaf nodes should be the code that's used the least.

## Now

![index](https://user-images.githubusercontent.com/537700/70938426-02a76a80-1ffb-11ea-9be7-3b1b32d7f362.png)

## Maybe future?

![maybe-future](https://user-images.githubusercontent.com/537700/70938671-76e20e00-1ffb-11ea-9560-99b5a73d0d79.png)

**Are you willing to submit a pull request to implement this change?** Yep.

christianbundy:
New hotness:

![Screenshot from 2019-12-26 20-47-15](https://user-images.githubusercontent.com/537700/71501753-f902db80-2820-11ea-98f0-3d35844a983c.png)

christianbundy:
cc: @mycognosist

mycognosist:
@christianbundy That's looking rad! I really like the ergonomics of the nav bar - no more unnecessary scrolling to reach the menu.

Have you come up with a solution for mobile?

christianbundy:
Thanks! Having to scroll to the top of the page always felt so silly, so a sidebar feels good. I was tempted to do a sticky header but:

`%ZyBgJ3rUgwKNY1VUO9CDgOY1MvODRrgfMoCEExynxm8=.sha256`

[...]

PR #95, opened 2020-01-27
Add simple nav sidebar on large screens

christianbundy:
Problem: Putting the navigation at the top of the screen makes it
impossible to use when you're scrolling through a page, which isn't a
good user experience. It was never meant to be permanent, and I think
everyone has pointed out that it's been a pain. See #28.

Solution: Super simple sidebar nav when people are on bigger screens.
This doesn't solve the problem on mobile, and it doesn't incorporate the
'popular' page's interval settings, but I think it's a step in the right
direction.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -456,3 +456,24 @@ input[type="search"] {
 hr {
   border: var(--pico) solid var(--fg-alt);
 }
+
+/* sidebar only appears on big screens */
+@media (min-width: calc(45rem)) {
+  body > nav > ul {
+    justify-content: right;
+    flex-direction: column;
+    margin-right: var(--kilo);
+    position: sticky;
+    top: var(--whole);
+  }
+  body > nav > ul > li {
+    margin-bottom: var(--whole);
+  }
+  main {
+    max-width: var(--measure);
+  }
+  body {
+    display: flex;
+    justify-content: center;
+  }
+}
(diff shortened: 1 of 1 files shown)


christianbundy:
Screenshot from 2020-01-27 12-36-15
Screenshot from 2020-01-27 12-35-57
Screenshot from 2020-01-27 12-35-40
Screenshot from 2020-01-27 12-35-22
Screenshot from 2020-01-27 12-35-06

christianbundy:
The merge history looks super janky but the weird sidebar movement problem is solved!

jedahan:
How do you feel about rebase and merge?

christianbundy:
> How do you feel about rebase and merge?

Happy to do it if you'd like, but not something I'd do recreationally. Usually I try to avoid rewriting Git history unless I *really* mess something up, like committing to the wrong branch, but if it's important to you I have no problem taking a few minutes and making my branch prettier.

jedahan:
Oh I didn’t mean rebase, I meant if I squashed on merge.

jedahan:
personally seeing 4 merge commits in a 7 commit history seems like unnecessary noise

christianbundy:
I'm conflicted -- on one hand, it'd be nice to have one commit per patch, but on the other it feels bad to rewrite the actual development history (especially with commit messages) and replace it with something simple. Feels like a 'hard truth vs simple lie' sort of thing.

If you think that's best I'm happy to follow your lead. Or we could discuss in #54 if you'd prefer? :man_shrugging: It's up to you, I trust your judgement here.

jedahan:
I'm going to try rebase and merge on this PR to see exactly how github does it, and then discuss how we should do it in the future.

Issue #98, opened 2020-01-27
Switch to capitalised first letters in menu


georgeowell:
**What's the problem you want to solved?**
Items in the menu and the submit button don't have the first letter capitalised. This kinda stands out seems unusual to me. 

**Is there a solution you'd like to recommend?**
Switching as general rule to capitalising the first letter unless there is a specific reason to have it that way.

christianbundy:
Thanks for bringing this up! Could you help pinpoint which menu you mean? There's both the navigation menu and the actions that you can do on each individual message. Or both?

Background: I used to have the message menu capitalized but I thought it looked weird and unbalanced. Lately I've been leaning toward all-caps actions, like how material design does it, but I haven't had time to experiment with it. It may be useful to know that lots of design cues have been influenced by news.ycombinator.com, for better or worse. :grimace

christianbundy:
Btw, I really appreciate the way that you identified a problem that you're having and suggested a small baby step in the right direction. There's tons of power in small iterative changes and it makes for a nice environment for writing and reviewing code.

georgeowell:
I think just sticking to a consistent pattern for this throughout. It would be great to get some icons up in there also. 

news.ycombinator.com is probably not the best thing to base our design from my perspective :sweat_smile:

christianbundy:
> I think just sticking to a consistent pattern for this throughout.

Yup! If you haven't already seen it, I'd love your feedback on #99.

> It would be great to get some icons up in there also.

Yep, at least some emoji would be great IMO.

georgeowell:
Closing due to https://github.com/fraction/oasis/pull/99

PR #99, opened 2020-01-27
Use capitalization for Oasis links and buttons

christianbundy:
Problem: Using all-lowercase-everything isn't really a standard around
the web and it might be better to use consistent capitalization. This
was brought up in: https://github.com/fraction/oasis/issues/98

Solution: This changes the main navigation to use links with the first
letter capitalized, like how Patchwork + Twitter + Mastodon + etc do it.
This means that we're consistently using sentence case everywhere, which
I think is our best option. Originally I tried experimenting with
all-caps for actions, which I found aesthetically pleasing, but you have
to reduce the font size to make it look good (bad!) and I was reading
that all-caps text is harder for friends with dyslexia or vision
impairments.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -46,6 +46,7 @@
   --common-radius: var(--micro);
   --measure: calc(var(--peta) + var(--mega));
   --line: 1.5rem;
+  --code-size: 85%;
 }
 
 * {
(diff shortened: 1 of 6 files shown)


christianbundy:
Screenshot!

Screenshot from 2020-01-27 15-56-20

(This screenshot has lots of other branches merged too)

cinnamon-bun:
Looks good!

I don't have time to try out the branch right now but the code looks ok.  Do we want Maintainers to test everything manually before merging it?

christianbundy:
> Do we want Maintainers to test everything manually before merging it?

I'm comfortable not doing this, just a quick glance at the code would be great. Really appreciate you noticing my backward `restart()` logic!

jedahan:
I can try out the branch right now.

christianbundy:
Thanks for the review and merge!

PR #101, opened 2020-01-28
Remove inline CSS to tighten CSP

christianbundy:
Problem: We were using inline CSS to show off the theme on the 'meta'
page, which meant that we couldn't tighten the CSP. This meant that
if you could figure out a CSS injection you'd be able to change how
Oasis looks, which could probably be malicious.

Solution: Add the CSS to `style.css` and remove that attack surface
area. Removing this inline CSS is a best practice, and while I don't
think a CSS injection is a super realistic threat there's no reason to
leave open holes in our CSP.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -456,3 +456,58 @@ input[type="search"] {
 hr {
   border: var(--pico) solid var(--fg-alt);
 }
+
+/* show off each theme color as a background */
+
+.theme-demo {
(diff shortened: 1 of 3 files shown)


christianbundy:
WOOPS plz do not merge

christianbundy:
I forgot that the thread indentation used inline CSS too, so I need to loosen the CSP here. In the future I think it should be simple to fix the indentation without inline CSS, but we probably don't want this merged right now.

PR #108, opened 2020-01-28
2.12.0

christianbundy:
**What's the problem you solved?** It's been two weeks since the latest release and there's been tons of improvements. Also I saw @georgeowell post a screenshot on SSB and realized that not everyone has all the new hotness. :sparkles: 

**What solution are you recommending?** Let's send it off! I want to do the changelog but don't have time right now, maybe tomorrow? I'm down. to release before adding the changelog, FWIW.

Once someone merges this I'll `npm publish`, but in the future we should make sure other maintainers have a way to release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@fraction/oasis",
-  "version": "2.11.0",
+  "version": "2.12.0",
   "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


christianbundy:
Actually I'm going to try to self-merge this just to see if I even *can* do a release. Bottoms up...

PR #120, opened 2020-01-30
Replace "must" with "shall" in contributing.md

christianbundy:
Problem: The word "shall" is ambiguous and seems pretentious rather than
friendly. More info here: https://www.plainlanguage.gov/guidelines/conversational/shall-and-must/

Solution: Replace "shall" with "must". It has the same semantics with
none of the ambiguity or cultural baggage.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -25,7 +25,7 @@ C4 provides a standard process for contributing, evaluating and discussing impro
 
 ## Language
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Btw I removed the 'require approvals' thing so we can just merge stuff now. :tada: It was only there so I didn't actually push to master, but since we require Travis to pass before merging I just turned it off. I think #119 is good to merge too?

christianbundy:
Woop woop, thanks for the merge!

PR #122, opened 2020-01-31
Make 'Popular' page faster

christianbundy:
Problem: The popular page is ungodly slow.

Solution: Make it faster! This is done by checking the timestamp before
the other constraints that we add to messages, which is mostly useful
because most messages that fail the filter will fail on the timestamp
check.

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `curl http://localhost:4515/public/popular/year` | 13.159 ± 0.480 | 12.843 | 14.450 | 1.21 ± 0.05 |
| `curl http://localhost:3000/public/popular/year` | 10.866 ± 0.163 | 10.594 | 11.133 | 1.00 |

### DIFF
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -685,11 +685,11 @@ module.exports = cooler => {
           source,
           pull.filter(msg => {
             return (
+              msg.value.timestamp > earliest &&
               typeof msg.value.content === "object" &&
               typeof msg.value.content.vote === "object" &&
               typeof msg.value.content.vote.link === "string" &&
-              typeof msg.value.content.vote.value === "number" &&
-              msg.value.timestamp > earliest
+              typeof msg.value.content.vote.value === "number"
             );
           }),
           pull.reduce(


jedahan:
are the other constraints necessary as well? would removing them speed things up?

christianbundy:
One way to find out! I did the thing that I thought made the most sense.

christianbundy:
(It's important that `link` is a string that points to a message and `value` is an integer that represents the vote. 1 is a like, 0 is an unlike)

cinnamon-bun:
I'm surprised this made such a dramatic speed difference!  I would have thought reading from disk would be the bottleneck.

The stream options are already filtering by `gt: earliest` - do we need to check again in the `pull.filter` or is that doing something different?

christianbundy:
Thanks for merging this!

It's unfortunately doing something different. :disappointed: The `message.timestamp` is the time that we received the message, added by FlumeDB and used by `gt`, but `message.value.timestamp` is the asserted timestamp that the author published. The `gt` uses the receive time to reduce the number of messages we have to process and the second check makes sure that it's *actually* posted from the correct time.

Issue #57, opened 2020-01-16
Don't show strangers in the main feeds


cinnamon-bun:
**What's the problem you want to solve?**
The Latest and Popular views seem to show everyone in my local database even if I'm not following them.  I'm usually only interested in threads started by, or participated in, by people I directly follow.

**Is there a solution you'd like to recommend?**
Small scale solution: limit those views to people I follow.

Here's the [query for Latest](https://github.com/fraction/oasis/blob/master/src/models/index.js#L585).  I don't know how to filter it by people I'm following (easier) or comments on threads started by people I'm following (trickier)

Large scale solution: If I was redesigning the navbar from scratch I would make the Gradient of Intimacy the first choice in the hierarchy, with filter options below it:

[Everyone]  [People you follow]  **[Mutual friends]**  [PMs]
...their [x] Posts   [ ] Comments  [ ] Likes
...sort by: Latest / Popular today / Popular this week

christianbundy:
Good catch, I think the 'large scale solution' is probably the right way to go. Thanks for pointing this out.

cinnamon-bun:
I'd love to eventually work towards this idea from Shellogram, but it's not the next minimum viable step :)

Users could add and edit their own views.  The expanded green box here lets you edit the view settings.  Underlined words open dropdowns to change that setting.

![Screen Shot 2020-01-16 at 3 44 41 PM](https://user-images.githubusercontent.com/32660718/72572391-37ffdc00-3877-11ea-96b0-978b4e5a8ad9.png)

There are so many settings that I found them overwhelming when they looked like dropdowns and checkboxes.  I used prose instead, hoping it would be easier to understand.

This kind of prose could be useful even if it's not editable, just explaining what's in a view (hidden behind a ❓ icon maybe)

georgeowell:
Very much into this!

christianbundy:
Resolved by #146

PR #123, opened 2020-01-31
Add basic following page

christianbundy:
Problem: We don't always want to see all of the messages on our
computer, sometimes we just want to see messages from the people we're
explicitly following. The 'Popular' and 'Latest' pages don't help with
that. See #57 

Solution: Add a super basic page that just shows the latest messages
from the people you're explicitly following.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -126,6 +126,11 @@ router
     };
     ctx.body = await publicLatest();
   })
+  .get("/public/latest/following", async ctx => {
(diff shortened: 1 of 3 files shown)


christianbundy:
Just a quick note: I don't feel strongly about this being separate from the 'Popular' or 'Latest' pages, but wanted to avoid mutating those pages until we could set the expectations for them more clearly and/or add filtering options.

PR #124, opened 2020-01-31
Specify exact Node.js engine support and test it

christianbundy:
Problem: Targeting only the latest version of Node.js and the latest LTS
means that we have a sort of rolling backward-compatibility that's hard
to reason about. If we break compatibility, we should release those
changes as a major version, but the current strategy makes it difficult
to tell exactly when we're breaking compatibility.

Solution: Specify the exact Node.js versions that we want to support in
`package.json` and add the specific tests to `.travis.yml`. In the
future we can break compatibility if it's helpful but we'll have to do
so explicitly and release the change as a major version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,14 @@
 # Use Node.js configuration.
 language: node_js
 
-# Both latest Node.js and latest LTS.
+# Test the latest and LTS releases >= LTS Dubnium (10).
 node_js:
-  - lts/*
+  - lts/dubnium
+  - lts/erbium
   - node
 
 # All of the operating systems!
-# Note: Windows support is very iffy.
+# Note: Windows support is iffy.
 os:
   - linux
   - osx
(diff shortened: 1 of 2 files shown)


jedahan:
What would the downsides be to changing `>= 12.0.0` to `^12.0.0`?

christianbundy:
That would add a warning for Node.js 13, which I think we want to support until Node 14 comes out. My internal heuristic is to support even-numbered [LTS] releases + the latest release, that way we're compatible with all of the LTS releases but we also get to keep compatibility with new changes coming down the pipeline.

Once Node 14 comes out, I'd imagine that we'd change `package.json` to:

```
^10.0.0 || ^12.0.0 || >=14.0.0
```

And once Node 15 comes out we'd change `.travis.yml` to:

```
  - lts/dubnium
  - lts/erbium
  - lts/foobar
  - node
```

PR #125, opened 2020-01-31
Update Flotilla to drop SSB-Links2

christianbundy:
Problem: We aren't using the index at all so Oasis is just wasting time
and CPU power.

Solution: Upgrade Flotilla and resolve the problem.


cinnamon-bun:
Is [ssb-links](https://www.npmjs.com/package/ssb-links) (aka links2) useful for us in the future?  The README has an example of indexing threads somehow, and I've been wondering how to make thread-centric views like Patchwork has.

christianbundy:
I'm not sure, I opened a discussion about this on SSB and got some useful replies. My understanding is something like "it should replace `links()` in SSB-DB" but I don't really understand why and I don't think there's any utility in having both. Happy to re-add it if it is, in fact, The Future, and if there's utility in having two links plugins at once.

```
%49X1n1MOoDo1gZRS4sgsa9pDL5wKCnIevgwMy8SISo4=.sha256
```

PR #127, opened 2020-01-31
Move Markdown handling from model module to view

christianbundy:
Problem: The model code was doing our Markdown rendering, which feels to
me like a layer violation because *generally* the model is meant to be a
thin abstraction over the underlying database without any concept of the
presentation layer.

Solution: Move the Markdown renderer to the view module and manage
Markdown rendering in the presentation layer.


cinnamon-bun:
Good idea!  Will merge once conflicts are resolved

PR #128, opened 2020-01-31
Fix SVG rendering by setting content-type

christianbundy:
Problem: Most browsers parse SVG files as XML and refuse to display it
in an `` tag. It's usually unsafe to have browsers try to sniff the
file type themselves, because they can be tricked into running unsafe
code, so we want to set the file type ourselves in the server.

Solution: Use the Is-SVG library for a quick-n-dirty check for whether a
buffer is an SVG. If so, we set the file type accordingly.


christianbundy:
Thank you for merging this!

Issue #117, opened 2020-01-29
Add profile picture and name to main menu (with link to profile)


georgeowell:
## What's the problem you want to solved?
Further replicate existing social media user layout expectations of homepage behaviour to display profile picture and name.

## Is there a solution you'd like to recommend?
* Add profile picture and name to main menu
* Profile picture and name link to own profile page when clicked
* Make rest of main menu items look nested with centre-ish alignment below that (I think that's how to describe it, see mock up)
* Remove existing "Profile" entry from the main menu

## Mock up
![Oasis Display Picture Menu mock up](https://user-images.githubusercontent.com/18296993/73322844-5d102980-423d-11ea-804f-a023a246d894.jpg)

## Notes
This will help distinguish accounts if multiple identities are implemented as mentioned in #4.

jedahan:
This looks like a nice improvement to me.

Riffing off the idea, what if the profile wasn't on the top of the list, but a divider, something like:

 Popular
 Latest
 Search
**:dog2: KawaiiPunk**
 Messages
 Mentions
 Preferences

Thinking behind this
- what if identity wasn't the most prominent thing
- highlights the boundaries of posts in relation to identity

cinnamon-bun:
+1 on putting it in the middle since that's where it goes on the Gradient of Intimacy as a divider between public and personal, e.g. https://github.com/fraction/oasis/issues/116#issuecomment-580425136

(This is an idea from the book A Pattern Language about human-centric buildings and spaces)

![](https://www.kerryveenstra.com/wp-content/uploads/2016/12/pattern-language-2.png)

https://books.google.com/books?id=hwAHmktpk5IC&printsec=frontcover&dq=a+pattern+language&hl=en&sa=X&ved=0ahUKEwiPvKWbjZbRAhUKymMKHagTCxwQ6AEIHDAA#v=onepage&q=%22127%20intimacy%20gradient%22&f=false

hng:
I've started working on this as it seems to be a good first issue. I would implement it as suggested by @jedahan

christianbundy:
Awesome, please reach out if you bump into any problems.

Issue #116, opened 2020-01-29
Rename "Inbox" to "Messages"


georgeowell:
**What's the problem you want to solved?**
I think "Inbox" is maybe the wrong term. It reminds people of work and email. Not chatting with other solarpunx. 

Coequally on Scuttleverse the posts are referred to as "Private Messages". Twitter and Facebook both refer to to this feature as Messages and I think we should do the same.

**Is there a solution you'd like to recommend?**
Rename "Inbox" to "Messages" in the main menu.

christianbundy:
Yes! Slightly concerned that 'messages' has a double-meaning since it refers to the payloads we're publishing via SSB in general. Do you think 'private' would be good or nah?

jedahan:
Private feels like a nice mirror of Public, and reinforces the security aspect. Both are improvements

cinnamon-bun:
In #57 we talked about adjusting the views to follow the Gradient of Intimacy, something like

Everyone
People you follow
Mutual friends
Mentions
Private

I think "Private" fits into this sequence a bit more clearly than "Messages".

christianbundy:
Resolved by #130, woo!

PR #130, opened 2020-02-01
Navbar refresh: reorder, rename, add emojis

cinnamon-bun:
In several issues we've discussed renaming pages in the navbar and changing their order.

This PR:
* Reorders the navbar links to follow the gradient of intimacy from public to private #57 #117
* Renames Inbox --> Private, and Meta --> Settings #116 to be more similar to Mastodon
* Adds emojis because there are now too many items in the navbar list to comprehend at a glance

Screen Shot 2020-01-31 at 4 08 59 PM

I'm not sure the emojis are a good idea.  If we don't like them after a couple of days we can remove them in another commit.  This is an accidental stress-test for our C4 process :)

I'd prefer to have simpler monochrome icons that are less distracting.  It was a challenge to find sensical emojis for these concepts.  I was especially trying to avoid these ones: 👤 👥 🆕 📈 👪

Also I didn't change the URLs for the pages yet; I'm going to wait for the dust to settle on this PR first.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -392,13 +392,15 @@ label {
 
 nav > ul {
   display: flex;
-  justify-content: space-between;
+  flex-wrap: wrap;
+  justify-content: flex-start;
   margin: 0;
   padding: 0;
 }
 
 nav > ul > li {
   list-style: none;
+  margin-right: var(--milli);
 }
 
 .profile {
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -247,7 +247,7 @@ exports.metaView = ({ status, peers, theme, themeNames }) => {
   return template(
     section(
       { class: "message" },
-      h1("Meta"),
+      h1("Settings"),
       p(
         "Check out ",
         a({ href: "/meta/readme" }, "the readme"),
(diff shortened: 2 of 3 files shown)


cinnamon-bun:
Hm, it looks bad on narrow screens:

Screen Shot 2020-01-31 at 4 28 37 PM

Since Latest was added in #123 there are too many items to go in a single row, and the emojis from this PR are wrapping strangely.

The emojis look weird but don't really break anything, but the navbar being too wide is affecting the scale of the rest of the page.

Fixing...

cinnamon-bun:
Now the navbar items wrap on narrow screens.  There are always at least 2 rows of them since there are so many.

They could probably have more vertical spacing for easy finger-poking, but this'll do for now

Screen Shot 2020-01-31 at 4 34 50 PM

christianbundy:
This is so great, thanks for putting this together!

Screenshot_2020-01-31 Oasis(1)
Screenshot_2020-01-31 Oasis

So friendly and inviting, I'm really excited to have this. :fireworks:

georgeowell:
Awesome! :champagne: :clinking_glasses:

PR #132, opened 2020-02-01
Remove follow button from your own profile

cinnamon-bun:
The "Follow" button was showing on your own profile.

I took a screenshot an hour ago and it wasn't there, and then it was.  Was there a change to capitalize the relationship strings ("You are following" etc)?  It would be good to refactor these relationship strings into constants.

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -50,9 +50,8 @@ exports.authorView = ({
 
   const contactFormType = areFollowing ? "Unfollow" : "Follow";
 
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
Ah I see we're moving development to our own forks to separate it from Maintainer work.  I'll do that next time.

christianbundy:
Thanks for fixing this, strongly agree about refactoring those into constants.j

PR #133, opened 2020-02-01
Make 'Following' page way faster

christianbundy:
Problem: The 'Following' page was super slow because we were doing a
MuxRPC request on every single message.

Solution: Do one request to see who we have relationships with, filter
out the people we aren't following, and then check against that list
instead of doing a bunch of MuxRPC calls.

|:---|---:|---:|---:|---:|
| `curl http://localhost:3000/public/latest/following` | 500.9 ± 163.4 | 392.1 | 881.4 | 1.00 |
| `curl http://localhost:3456/public/latest/following` | 4663.7 ± 184.6 | 4438.6 | 5075.2 | 9.31 ± 3.06 |

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -649,6 +649,15 @@ module.exports = cooler => {
     },
     latestFollowing: async () => {
       const ssb = await cooler.connect();
+      const { id } = ssb;
+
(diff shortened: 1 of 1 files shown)


christianbundy:
Thank you!

PR #139, opened 2020-02-01
Add basic scaffolding for internationalization

### LINKED ISSUE #4: identities, language, etc
I think it's important that new clients embrace new ssb abilities from the get go. For example, we can now do multiple identities quite easily (using `ssb-identities`) but most clients assume that there is only a single identity.

In yap, I implemented this using cookies - the reason being that it makes the url generic - you could share a link if you wanted to, and it would make sense without knowing the selected user.
(not that you really want to be able to share links, but it seemed like clean design)

I planned the same approach for language. basically, you do a post, which sets the cookie. you just need an ordinary form on the front end.

What do you think? I'm keen to have a back end that can work with oasis and yap.

christianbundy:
Problem: It was impossible to do any internationalization because
strings were all embedded throughout Oasis.

Solution: Add an internationalization submodule that provides
language-specific strings for the text elements in views. In future
commits we can add language selection and fallbacks for when the
selected language doesn't support the text we need to have translated. Resolves #4


cinnamon-bun:
🌟 I was just thinking about this!  Awesome

christianbundy:
Related: #128 

I tried to add identities earlier but bumped into https://github.com/ssbc/ssb-db/issues/282 which is a blocker.

cinnamon-bun:
I'm going to hold off on merging anything else until this one gets in, since it's big and it will probably keep having conflicts

christianbundy:
Sounds good, thanks for all the time you've spent merging things! :heart:

PR #140, opened 2020-02-02
Fix help link in readme

### LINKED ISSUE #56: Help link in readme is broken
**What's the problem you want to solved?** The "help" link in the readme points to:

https://github.com/fraction/oasis/issues/new/choose

This page is broken.

**Is there a solution you'd like to recommend?** Instead, I think it should point to:

https://github.com/fraction/oasis/issues/new/

christianbundy:
Problem: Help link in readme was pointing to the issue type selection
which is no longer useful because we just have one template. Resolves #56 

Solution: Fix the link and point directly to the issue creation link.


christianbundy:
Thanks a bunch for merging this!

Issue #107, opened 2020-01-28
Search is broken when the query is < 3 characters


christianbundy:
**What's the problem you want to solved?** The search page returns all results at once, with a maximum of 64 results, but since they're all returned at once the search is super broken any time that you search for something uncommon. If there are < 64 results, I think the current implementation is just sitting there waiting for more results? I'm not 100% sure.

**Is there a solution you'd like to recommend?** We should probably be streaming results everywhere instead of collecting them in memory and returning them all at once.

christianbundy:
F7U12. The way we're doing templates does *not* seem compatible with streaming HTML interfaces. This is a good thing to learn early in the project, and it's something that we'll likely have to fix. For maximum performance, we want the views to be able to export an HTML stream rather than entire blocks of code. My intuition is that this will make Oasis *very* snappy but might need to be well-documented so it isn't inaccessible for new contributors.

christianbundy:
Nope, SSB-Search is just broken. https://github.com/ssbc/ssb-search/issues/8

cinnamon-bun:
Is this resolved by #141 ?

christianbundy:
Yup, closing now!

PR #141, opened 2020-02-02
Fix search hanging when querying 3 or fewer chars

christianbundy:
Problem: SSB-Search has a bug where too few characters just hangs the
search indefinitely and never returns. https://github.com/fraction/oasis/issues/107

Solution: Enforce a minimum length of 3 characters in the search. I
bumped into another bug where HyperScript, a dependency of HyperAxe,
doesn't support the `minlength` attribute, so I had to deploy a small
workaround for that too. The fixes aren't very pretty but they're better
than just ignoring the problem.


cinnamon-bun:
So many fixes!  Great work

✨ c o d e ✨  q u a l i t y ✨

PR #145, opened 2020-02-02
Make "key not found" error more useful

### LINKED ISSUE #144: Key not found in database
**What's the problem you want solved?** If you're missing a message and try to go to the thread then it throws a crappy error message.

```
Error: non-error thrown: {"message":"Key not found in database [%u8t50zvXB/Tj3TTSFxgwKxz6zZS0hyYGpITFSZGL+Q0=.sha256]","name":"NotFoundError","stack":"NotFoundError: Key not found in database [%u8t50zvXB/Tj3TTSFxgwKxz6zZS0hyYGpITFSZGL+Q0=.sha256]\n    at /home/christianbundy/src/fraction/oasis/node_modules/levelup/lib/levelup.js:192:15\n    at /home/christianbundy/src/fraction/oasis/node_modules/encoding-down/index.js:75:21"}
    at Object.onerror (/home/christianbundy/src/fraction/oasis/node_modules/koa/lib/context.js:113:40)
    at onerror (/home/christianbundy/src/fraction/oasis/node_modules/koa/lib/application.js:163:32)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
```

I'm seeing this on:

- http://localhost:3000/thread/%25u8t50zvXB%2FTj3TTSFxgwKxz6zZS0hyYGpITFSZGL%2BQ0%3D.sha256
- http://localhost:3000/thread/%25EDeBAjTfkOGGszsSlTLiO7HRCGY1aPTyjKduii2xOMw%3D.sha256

We should be catching this error and providing a more helpful error message.  Even just "This message isn't available, Oasis will look around for it so maybe check back later. You've done nothing wrong" or something would be way better than blowing up..

christianbundy:
Problem: If the user tries to see a thread and the link points to a
message we don't have, then we don't have any way to display anything in
the thread. How could we even know which thread it's in? Resolves #144 

Solution: Throw the error but make it more useful and fix the "non-error
thrown" verbiage that we've inherited from a dependency trying to throw
a non-error.

---

If this diff looks scary, try removing the whitespace. We have to do this because of the weird non-error being thrown by a deeper dependency. :man_facepalming:

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -811,187 +811,195 @@ module.exports = cooler => {
       const myFeedId = ssb.id;
 
       const options = configure({ id: msgId }, customOptions);
-      const rawMsg = await cooler.get(ssb.get, options);
+      return cooler
+        .get(ssb.get, options)
+        .then(async rawMsg => {
+          debug("got raw message");
 
-      debug("got raw message");
+          const parents = [];
 
-      const parents = [];
+          const getRootAncestor = msg =>
+            new Promise((resolve, reject) => {
+              if (msg.key == null) {
+                debug("something is very wrong, we used `{ meta: true }`");
+                resolve(parents);
+              } else {
+                debug("getting root ancestor of %s", msg.key);
 
-      const getRootAncestor = msg =>
-        new Promise((resolve, reject) => {
-          if (msg.key == null) {
-            debug("something is very wrong, we used `{ meta: true }`");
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
🌟 for friendly error messages

PR #146, opened 2020-02-03
No strangers

christianbundy:
Problem: We shouldn't be showing any strangers on the popular or latest
pages. We shouldn't be showing anyone who's been blocked on any of those
feeds.

Solution: Don't show blocked people on any pages and don't show
strangers on the popular / latest pages.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -56,40 +56,6 @@ module.exports = cooler => {
         dest: feedId
       });
       return raw;
-    },
-    all: async feedId => {
-      const ssb = await cooler.connect();
-      const raw = await cooler.read(ssb.about.read, {
-        dest: feedId
-      });
-
-      return new Promise((resolve, reject) =>
-        pull(
-          raw,
-          pull.filter(message => message.value.author === feedId),
-          pull.reduce(
-            (acc, cur) => {
-              const metaKeys = ["type", "about"];
-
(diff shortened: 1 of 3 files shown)


cinnamon-bun:
Resolves #57 Don't show strangers

I see it's filtering out "me".  Interesting, I never noticed if I was included in the main feeds in Patchwork or not.  I think it feels a bit friendlier to include yourself in the feeds so you can see yourself amongst your friends?

christianbundy:
That makes sense, thank you! I'd either merge a patch from you (either set the default to `me = true` or change `{ me: true }` where we want it) or I'll push a patch to fix in the next few days.

src/models.js
```diff
@@ -304,6 +272,34 @@ module.exports = cooler => {
     });
   };
 
+  const socialFilter = async ({
+    following = null,
+    blocking = false,
+    me = false
+  } = {}) => {
```
cinnamon-bun: A comment could be nice here to describe that `null` means don't apply a filter, `true`/`false` mean apply a filter for that value respectively?

Also if omitted when calling the function they get these as default values?  So the function doesn't need to deal with them being undefined?

(I think this makes sense, I'm just not familiar with unpacking an object and setting defaults in the function signature like this.  I don't know what the `= {}` is doing on 279)

Issue #149, opened 2020-02-03
Should we advertise a bleeding-edge install from master?


christianbundy:
**What's the problem you want solved?** We're making lots of improvements really fast and I'd like to let people try them without building from source. Making new versions all the time is a pain, and stretches our maintenance even thinner.

**Is there a solution you'd like to recommend?** Should we add an `npm --global install github:fraction/oasis` option to the readme for people who want the bleeding edge? Or is that overly complicated? If we do that, people can just upgrade and it'll pull the latest version from GitHub without forcing us to make a release every day.

PR #151, opened 2020-02-03
Add bleeding-edge install instructions to readme

christianbundy:
Problem: https://github.com/fraction/oasis/issues/149

Solution: Add note to the readme advertising the npm install straight
from GitHub, which will pull the latest `master`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,19 @@ Options:
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


christianbundy:
I'd merge a patch to move it to `install.md`. I put it here because I think most people using the repo would probably want the bleeding-edge install, but I wouldn't mind putting it in the `install.md` instead.

PR #154, opened 2020-02-03
Add invite code

### LINKED ISSUE #97: Oasis user onboarding stratergy
Me again :smile: 

Scuttlebutt clients in general don't make it easy to onboard people. I think this should be one of the main focuses of Oasis (along with awesome user safety features #93). 

Support for ssb-rooms would be handy. @arj03 has also been working on some really cool stuff with ssb-browser. I wonder if we can leverage some of these technologies. Here is [a demo](https://between-two-worlds.dk) of ssb-browser.

Here are some rough ideas of how to improve the experiance over previous clients:

## SSB Rooms 
Support for SSB room would help new users connect remotely or in different network conditions.

https://github.com/staltz/ssb-room/
https://www.manyver.se/blog/announcing-ssb-rooms

## Partial replication
It's a real pain for new users to have to wait 15 minutes to do the initial sync after making their first connections on the Scuttleverse. 

> Partial replication is done by only getting the last few post messages for users as a way of onboarding users quickly. The client can download and index around 5.000 messages in 30 seconds on my really slow laptop (roughly same speed on a phone). For this, a blob (generate-onboarding-json.js) must be provided that serves as a trusted onboard mechanism.

> This will provide an initial load of data and people. You then sync the feeds of these people, but selective sync can be done by following people directly.

> The ssb-partial-replication plugin is needed for faster sync. Also the ssb-get-thread plugin is used for browsing threads you don't currently have, such as from people outside the people you have synced or older messages because of the partial replication nature.

https://github.com/arj03/ssb-browser-demo
https://github.com/arj03/ssb-partial-replication
https://github.com/arj03/ssb-get-thread
https://github.com/ssbc/ssb-contact-msg

## Peer invites
A way to help bring people onto the Scuttleverse with a personalised invite.

> Peer invites is another way to onboard people. The client allows one to generate and use peer invites. The invites can include an list of people to follow beside the user that created the invite. In doing so, these invites serves the same function as the onboarding blob.

https://github.com/arj03/ssb-browser-demo#peer-invites

<strike> ## Onboarding blob
A way to help people follow all the people you are following or some combination of such.

https://github.com/arj03/ssb-browser-demo#onboarding-file</strike>

christianbundy:
Problem: There was no way to onboard new users since we couldn't redeem
invites.

Solution: Add basic follow-back invites to the settings page. This takes
an invite string, runs it through invite.accept, and either returns the
error in full *or* redeems the invite quietly. Resolves #97 

---

Note: This branch contains commits from the other PRs for i18n and such. Once the other PRs are merged I'll squash this and mark it 'Ready for review'.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -560,6 +560,11 @@ router
   .post("/meta/conn/restart", koaBody(), async ctx => {
     await meta.connRestart();
     ctx.redirect("/meta");
+  })
+  .post("/meta/invite/accept", koaBody(), async ctx => {
+    const invite = String(ctx.request.body.invite);
+    await meta.acceptInvite(invite);
+    ctx.redirect("/meta");
   });
 
 const { host } = config;
--- a/src/models.js
+++ b/src/models.js
@@ -193,6 +193,10 @@ module.exports = cooler => {
     connRestart: async () => {
       await models.meta.connStop();
       await models.meta.connStart();
+    },
+    acceptInvite: async invite => {
+      const ssb = await cooler.connect();
+      return await cooler.get(ssb.invite.accept, invite);
     }
   };
 
--- a/src/views/i18n.js
+++ b/src/views/i18n.js
@@ -3,9 +3,9 @@ const { a, strong } = require("hyperaxe");
 module.exports = {
   en: {
     // navbar items
-    extended: "Extended",
     extendedDescription:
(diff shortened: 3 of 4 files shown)


cinnamon-bun:
Thanks for adding this!  I'm not familiar with peer invites yet but agree it's important for onboarding

PR #157, opened 2020-02-04
Add description of each feed at the top of the page

cinnamon-bun:
### Problem
We have a lot of feeds with short names (Topics, Extended...) but it's not clear what they really mean.

### Solution
This adds a title and descriptive text to most of the feeds:

Screen Shot 2020-02-04 at 2 01 55 PM

Added to:
* [x] Extended
* [x] Popular
* [x] Latest
* [x] Topics
* [x] An author's Likes page

Not changed
* Author profile page (it already has an info card about the author)
* Search (it already has the search box)

### Future work for another day

Add the remaining pages:
* [ ] Mentions
* [ ] Private

Other things
* [ ] Improve the CSS - I'm not especially happy with the layout and fonts
* [ ] I embedded `\n` into the i18n strings and used CSS `whitespace: pre-line` to make it retain newlines.  This feels weird.
* [ ] Highlight the current page in the navbar so we don't need to repeat the page name in an `h1`
* [ ] Put the author's profile image on the Likes page
* [ ] Centralize generation of URLs

I also rearranged the i18n strings into a more logical order.  That should have been a separate patch, sorry :)

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -321,6 +321,11 @@ section {
   box-sizing: border-box;
 }
 
+.viewInfo {
+  background: none;
+  white-space: pre-line; /* preserve newlines */
+}
+
 section > header {
   height: var(--line);
   display: flex;
--- a/src/index.js
+++ b/src/index.js
@@ -48,6 +48,7 @@ const {
   commentView,
   extendedView,
   latestView,
+  likesView,
   listView,
   markdownView,
   metaView,
@@ -360,10 +361,14 @@ router
   })
   .get("/likes/:feed", async ctx => {
     const { feed } = ctx.params;
-
     const likes = async ({ feed }) => {
-      const messages = await post.likes({ feed });
-      return listView({ messages });
+      const pendingMessages = post.likes({ feed });
+      const pendingName = about.name(feed);
+      return likesView({
+        messages: await pendingMessages,
+        feed,
+        name: await pendingName
+      });
     };
     ctx.body = await likes({ feed });
   })
--- a/src/index.js
+++ b/src/index.js
@@ -46,13 +46,16 @@ const { about, blob, friend, meta, post, vote } = require("./models")(cooler);
 const {
   authorView,
   commentView,
+  extendedView,
+  latestView,
   listView,
   markdownView,
   metaView,
-  publicView,
+  popularView,
   replyView,
   searchView,
-  setLanguage
+  setLanguage,
+  topicsView
 } = require("./views");
 
 let sharp;
@@ -114,7 +117,7 @@ router
         ul(option("Day"), option("Week"), option("Month"), option("Year"))
       );
 
-      return publicView({
+      return popularView({
         messages,
         prefix
       });
@@ -123,15 +126,15 @@ router
   })
   .get("/public/latest", async ctx => {
     const messages = await post.latest();
-    ctx.body = await publicView({ messages });
+    ctx.body = await latestView({ messages });
   })
   .get("/public/latest/extended", async ctx => {
     const messages = await post.latestExtended();
-    ctx.body = await publicView({ messages });
+    ctx.body = await extendedView({ messages });
   })
   .get("/public/latest/topics", async ctx => {
     const messages = await post.latestTopics();
-    ctx.body = await publicView({ messages });
+    ctx.body = await topicsView({ messages });
   })
   .get("/author/:feed", async ctx => {
     const { feed } = ctx.params;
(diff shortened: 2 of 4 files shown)


cinnamon-bun:
Here are the descriptions.  I inspected all the model code to figure this out.  Is it correct?

### Extended
Who: People you're not following
What: Posts and comments

### Popular
Posts and comments (from anyone) with the most hearts (from people you follow)

### Latest
Who: People you're following
What: Posts and comments

### Topics
Who: People you're following
What: Posts. (To see comments, click the timestamp of a post)

### @ cinnamon's likes
_(there's no description on the likes view)_

src/assets/style.css
```diff
@@ -321,6 +321,11 @@ section {
   box-sizing: border-box;
 }
 
+.viewInfo {
```
cinnamon-bun: This is used on a <section> to turn off the background color for the new info box

src/index.js
```diff
@@ -123,15 +127,15 @@ router
   })
   .get("/public/latest", async ctx => {
     const messages = await post.latest();
-    ctx.body = await publicView({ messages });
+    ctx.body = await latestView({ messages });
```
cinnamon-bun: There wasn't any access to `i18n` in the main index file, so I couldn't grab the description for each view and just pass it into the generic `publicView()`.  Instead I made separate view functions for each specific view in `views/index.js` where I could grab the i18n strings.  Not too elegant to make so many new functions like this, but probably best to keep i18n confined in the view side.

src/index.js
```diff
@@ -357,10 +361,14 @@ router
...
     const likes = async ({ feed }) => {
-      const messages = await post.likes({ feed });
-      return listView({ messages });
+      const pendingMessages = post.likes({ feed });
+      const pendingName = about.name(feed);
+      return likesView({
+        messages: await pendingMessages,
+        feed,
+        name: await pendingName
+      });
```
cinnamon-bun: Don't need to use Promise.all - this is another way to run the promises in parallel

src/assets/style.css
```diff
@@ -321,6 +321,11 @@ section {
...
+  background: none;
+  white-space: pre-line; /* preserve newlines */
+}
+
```
cinnamon-bun: The new info boxes are `<section>`s with the background fill turned off

src/index.js
cinnamon-bun: This is a way to run promises in parallel without using Promise.all

src/views/index.js
```diff
@@ -515,10 +516,41 @@ exports.metaView = ({ status, peers, theme, themeNames }) => {
...
+    viewTitle ? h1(viewTitle) : null,
+    viewDescription ? em(viewDescription) : null
+  );
+};
+
+exports.likesView = async ({ messages, feed, name }) => {
+  const authorLink = a(
+    { href: `/author/${encodeURIComponent(feed)}` },
+    "@" + name
+  );
```
cinnamon-bun: It would be good to make some functions to generate common URLs like this

src/index.js
christianbundy: 🤯

christianbundy:
Thanks for this patch, looks great! :heart: No rush on the other pages, any explanation is an improvement.

src/index.js
christianbundy: @cinnamon-bun I just tried this in another project and it doesn't seem like it's doing the `Promise.all()` thing. Am I doing it wrong?

```javascript
const crypto = require("crypto");

const bullshit = () =>
  new Promise(resolve => {
    const id = crypto.randomBytes(8).toString("hex");
    console.log(`start ${id}`);
    const timeout = (Math.random() + 1) * 1000; // 1-2 seconds
    setTimeout(() => {
      console.log(`end ${id}`);
      resolve();
    }, timeout);
  });

const main = async () => {
  const results = {
    x: await bullshit(), // takes 1-2 seconds
    y: await bullshit(), // takes 2-4 seconds
    z: await bullshit()  // takes 3-6 seconds
  };
};

main();
```

cinnamon-bun: It works by separating things into two phases:
1. generate some promises so they start doing work in the background, but don't await them yet
2. await the promises

Like:
```js
const main = async () => {
  // PHASE 1
  // start them all in parallel but don't wait for them to finish yet
  const bs1promise = bullshit();
  const bs2promise = bullshit();
  const bs3promise = bullshit();

  // PHASE 2
  // this large expression will block until all the promises inside it are ready
  const results = {
    x: await bs1promise,
    y: await bs2promise,
    z: await bs3promise
  };
};
```

In your example I think the laungage doesn't take advantage of the obvious lack of dependencies between x, y, and z, it just scans linearly through the syntax, so we get stuck `await`ing the first one and it hasn't started the rest yet.

christianbundy: Oh, that makes sense -- thanks for explaining that to me! It's very clever.

Issue #160, opened 2020-02-05
Publish shouldn't be at the top of every page


christianbundy:
**What's the problem you want solved?** Publish being on the top of all the public pages takes up lots of screen space and 99% of the time I don't want to actually publish anything. I don't really like having to scroll every time I go to a new page.

**Is there a solution you'd like to recommend?** Move publish to its own page, 'Publish'?

georgeowell:
Ah I was going to post an issue about this. 

What if it was a dropdown with the little arrow. It could be just a box saying publish that could be expanding. It would take up like 1/8th of the space.

However there is an issue that arises here that suggests the new Publish page that you have already committed is the way to go.

Currently it may give users the impression that they are only posting to the specific feed they are viewing (i.e. only posting to followers or only posting to the extended network) which isn't true. A general publish page for public messages may be the way to go.

jedahan:
I really like the idea where any place we are about to write something we could share what context it will be seen. Reference to facebooks drop down globe/people icon for public vs friends

jedahan:
I’m also a fan of having the drop down on all pages, encouraging more active contribution vs passive consumption, though that might be reading into it a little too much. See apps that have a floating “+” action

PR #161, opened 2020-02-05
Add view info for Private, Mentions, and Publish

christianbundy:
Problem: The Private and Mentions page didn't have view labels yet, and
Publish should be its own page instead of being at the top of every
page.

Solution: Inspired by @cinnamon-bun's work to add friendly view labels,
plus a new Publish page.  This also moves the period selection from the
popular page into the view label, which felt better to me with the
previous `<section>` background. I also tried a different text format
for describing the pages, using a common form and using `<strong>` to
draw attention to any change from "Posts from people you follow, sorted
by recency" which feels like the expected default for most people.

@cinnamon-bun: To me this feels like a fun back-and-forth where I'm
riffing on your work and hoping that you do the same, but if it feels
wrong/rude please let me know. Trying to work on designs with a system
like C4 is super new for me and I wouldn't be surprised if there are
pain points to fix!

Resolves https://github.com/fraction/oasis/issues/160
Resolves https://github.com/fraction/oasis/issues/163

Screenshot_2020-02-04 Oasis

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -321,11 +321,6 @@ section {
   box-sizing: border-box;
 }
 
-.viewInfo {
-  background: none;
-  white-space: pre-line; /* preserve newlines */
-}
-
 section > header {
   height: var(--line);
   display: flex;
@@ -398,7 +393,7 @@ label {
 nav > ul {
   display: flex;
   flex-wrap: wrap;
-  justify-content: space-around;
+  justify-content: space-between;
   margin: 0;
   padding: 0;
 }
--- a/src/index.js
+++ b/src/index.js
@@ -51,8 +51,11 @@ const {
   likesView,
   listView,
   markdownView,
+  mentionsView,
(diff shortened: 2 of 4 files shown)


cinnamon-bun:
Thanks!  I made a rough proof of concept because I wanted it to happen, and now it looks much better and more consistent after your changes.  🌟

I had a vague aesthetic opinion that the info box was better if it looked different than a regular post box, but removing the background was not a good solution because it messed up the visual hierarchy of the page.  Maybe a different background color for all of the system boxes versus post boxes?  In any case this patch is an improvement on top of mine, so thank you!

This also lays the groundwork for adding more options to a view, if we want to try that later (what kinds of messages? who? sort order, filtering, etc)

cinnamon-bun:
I guess now every view follows the same pattern of a control panel box followed by a bunch of content boxes (except Settings has no content boxes)

Issue #156, opened 2020-02-04
New socialFilter change hides your messages when replying to thread


christianbundy:
**What's the problem you want solved?** The new socialFilter() that needs to be documented (#155) caused a bug that makes it so that when you hit "Comment" and look at the thread all of your previous comments are missing. :scream: 

**Is there a solution you'd like to recommend?** Set the default `{ me = null }` instead of `{ me = false }`

PR #162, opened 2020-02-05
Fix socialFilter() to change `{ me = null }`

christianbundy:
Problem: The socialFilter was hiding posts published by the user, which
felt weird and uncanny.

Solution: Fix the default so that `{ me }` isn't hidden from a view
unless the model specifically wants that to happen.

Resolves https://github.com/fraction/oasis/issues/156


christianbundy:
Thanks for merging!

Issue #153, opened 2020-02-03
Some emojis are being rendered with the default font


christianbundy:
**What's the problem you want solved?** If you have a font that supports tons of code-points, sometimes it'll render things that we want displayed as emojis. For example, the 'Settings' gear renders as text for me. This is inconsistent with other emoji and we *really* want ever emoji rendered the same way.

**Is there a solution you'd like to recommend?**

```css
span.emoji {
  font-family: emoji;
}
```

And then wrap all characters that we want displayed as emoji in this class. This will force the emoji font to be used, falling back to the text font, instead of the current situation where it uses the text font when it can and falls back to the emoji font. Bonus points for adding SSB-Markdown integration [like this](https://github.com/ssbc/patchwork/blob/c76136fb1628ec7fd4bb10b181d70ced01bb120c/lib/depject/message/html/markdown.js#L49-L56).

Issue #150, opened 2020-02-03
Linebreak between nav emoji and page name


christianbundy:
**What's the problem you want solved?** At a *very* specific width I'm seeing the nav links break into two lines, which I don't think we want.

![Screenshot_2020-02-03 Oasis](https://user-images.githubusercontent.com/537700/73674607-bde89900-4665-11ea-8693-b8261d6834a2.png)

**Is there a solution you'd like to recommend?** I think we should separate the emoji and the page name with a non-breaking space (`&nbsp;` in HTML, `\xa0` in JS).

PR #168, opened 2020-02-05
Fix emoji and breaking space in main nav

christianbundy:
Problem: The regular space in the sidebar emoji was breaking the line at
a specific viewport width and the emoji were being shown as the wrong
font.

Solution: Use a non-breaking space and `font-family: initial` for
full-color emoji instead of using the system font.

Resolves https://github.com/fraction/oasis/issues/150
Resolves https://github.com/fraction/oasis/issues/153

image

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -491,3 +491,11 @@ hr {
     max-width: none;
   }
 }
+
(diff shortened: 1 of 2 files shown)


cinnamon-bun:
I love that our screenshots all use different color themes and emoji fonts.

Issue #138, opened 2020-02-01
Content-Type for blob images


cinnamon-bun:
Opening an image in a new tab doesn't work well - the browser renders it as text.

I guess we need to figure out the file type on the server and set the Content-Type header.  At least for some common image and video types.

**Prior work**
This has already been done for SVG with #128 

Images are normally served from `/blob/...hash...`.  There's another endpoint, `/image/64/...hash...`.  If `sharp` is installed that resizes and crops images, otherwise it serves them raw.  It always sets Content-Type to `png` no matter the original image type, but browsers seem to do ok even if the data is actually JPG.

**Security implications**

???

christianbundy:
> Opening an image in a new tab doesn't work well - the browser renders it as text.

Are you noticing this problem on `/blob` or `/image` URLs? Are there specific images that aren't working? I don't think that I've noticed this problem [yet].

> It always sets Content-Type to png no matter the original image type, but browsers seem to do ok even if the data is actually JPG.

I think this made sense when Sharp was resizing them, but since Sharp is optional we should double-check that this is only happening if Sharp has turned them into a PNG.

christianbundy:
Found the underlying bug I think you're describing. The broken images you were seeing was actually your browser being shown text and being told it was a PNG. :man_facepalming: Here's what the error looks like if your browser shows it as text:

```
TypeError: fakeImage(...).then is not a function
    at /home/christianbundy/src/fraction/oasis/src/index.js:285:32
    at new Promise (<anonymous>)
    at image (/home/christianbundy/src/fraction/oasis/src/index.js:282:14)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
    at async /home/christianbundy/src/fraction/oasis/src/index.js:316:16
    at async /home/christianbundy/src/fraction/oasis/src/http.js:28:5
    at async /home/christianbundy/src/fraction/oasis/node_modules/koa-mount/index.js:52:26 {
  expose: true
}
```

cinnamon-bun:
My bug description was kind of meandering and led to another different bug :D 

I meant - find an image someone has posted (not an avatar), right-click the image and open in a new tab.

Example: this post

[...]

PR #169, opened 2020-02-06
Remove nosniff from blob URLs

christianbundy:
Problem: We use nosniff to keep the web browser from getting confused
about what kinds of content we're serving in Oasis, but this causes
problems for blob URLs that have arbitrary data.

Solution: Remove nosniff on blob URLs to let the browser figure out what
kind of content we're serving.

Resolves https://github.com/fraction/oasis/issues/138

Now works in Chromium!

Screenshot from 2020-02-05 17-02-28


cinnamon-bun:
Travis randomly failed on Windows because it couldn't install sharp.  Restarting the job...

cinnamon-bun:
Thanks for fixing this!

Issue #48, opened 2020-01-09
Public viewer


christianbundy:
**What's the problem you want to solved?** It's fun to run a pub + viewer but hard to do with Oasis. I have a branch called `public`, but it's not up-to-date with `master`.

**Is there a solution you'd like to recommend?** Maybe we should have support for Oasis as a public viewer? Would anyone want to use this?

christianbundy:
Now that the demo is outdated, this is more important than ever. :~)

cinnamon-bun:
Features needed for this (please let me know if I misinterpreted)

- [ ] Add command line flag for "public viewer" mode, which:
  - [ ] Makes everything read-only
  - [ ] Hides feeds which haven't opted in to public viewing
  - [ ] Sets more cache-y cache headers (in case the server gets a lot of traffic)
  - [ ] ❓robots.txt to exclude search engines?
- [ ] Add a button to your profile to opt in or out of public viewing by publishing a message to your feed; also show your opt-in status there.  (Default is opt-out.)

What if there was also `blog mode` - a public viewer that's focused around one person's feed, so you could put your own content online like a blog.  It would be limited to just your profile, maybe even just your own root-level posts with no comments to minimize privacy issues.

I'm wary of public viewers; people tend to quote each other in messages without realizing they might be publishing the content they quote to a public viewer if they themselves have opted in.
Mitigations:
- [ ] One-person public viewer which doesn't show comments, as above
- [ ] If you've opted in to a public viewer, make this obvious when you're writing a message or looking at your own profile
- [ ] Show if other people have opted in on their profile pages (it can be less obvious there)

christianbundy:
Thanks for speccing this out! That's really useful, and it's especially the kind of work that I have a hard time with sometimes. :heart: 

> I'm wary of public viewers; people tend to quote each other in messages without realizing they might be publishing the content they quote to a public viewer if they themselves have opted in.

I'm very conflicted about `publicWebHosting` in general. I've talked about this a few times on SSB, but I have lots of anxiety around someone mistaking the `publicWebHosting` option for any sort of real privacy, and I'm scared that we're normalizing this "you're safe here" feeling when posting something permanently on the internet.

I have this gut feeling that public viewers help bridge that gap, and make our lack of privacy harder to ignore, but I'm also open to the idea that my gut feeling is wrong and that public viewers like oasis-demo are actively harmful or dangerous.

[...]

Issue #110, opened 2020-01-28
The likes page shows stuff that's not liked anymore


Powersource:
The page mostly works fine for me but this message that I liked and then unliked (tried to look at who had liked it :P ) still shows up, and twice. I'm guessing it's just displaying every type 'vote' message in my log.

![image](https://user-images.githubusercontent.com/1427063/73308151-efa1d000-421f-11ea-9c32-eaadbffbb9b3.png)

christianbundy:
Nice catch! Really appreciate this bug report. For whoever fixes this, here's the relevant code:

https://github.com/fraction/oasis/blob/82f080f3c6b034ae38eba8a62b04de425b5d088e/src/models/index.js#L536-L590

And here's the correct vote-counting code that's used to display the number of likes on a post:

https://github.com/fraction/oasis/blob/82f080f3c6b034ae38eba8a62b04de425b5d088e/src/models/index.js#L335-L363

Maybe the correct behavior should be abstracted into a function and used in both spots?

cbleslie:
@christianbundy -  Nicety? Shipping with this might cause frustration, but not going to destroy a users experience?

PR #181, opened 2020-02-08
Add --public option for public viewers

christianbundy:
Problem: It's hard to show off Oasis or take screenshots without
respecting the `publicWebHosting` convention. While `publicWebHosting`
lacks a formal specification and I'm a bit confused about what its
boundaries are, it sounds like some of our friends would like to avoid
us publishing any of their content on the public web if we can avoid it.

Solution: Add --public option that turns Oasis into a public web viewer.
This makes it **slightly inconvenient** to see these public posts, but
should absolutely not be mistaken for a privacy guarantee. Only HTTP GET
endpoints are allowed, so random people can't publish or change
settings. The name, avatar, description, content warning, and message
contents are replaced with "Redacted", but again, this is all public
information that we can never provide real privacy for.

Resolves https://github.com/fraction/oasis/issues/48

Screenshot_2020-02-08 Oasis(1)

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -31,6 +31,12 @@ module.exports = () =>
       default: 3000,
       type: "number"
     })
+    .options("public", {
+      describe:
+        "Assume Oasis is being hosted publicly, disable HTTP POST and redact messages from people who haven't given consent for public web hosting.",
+      default: false,
+      type: "boolean"
+    })
     .options("debug", {
       describe: "Use verbose output for debugging",
       default: false,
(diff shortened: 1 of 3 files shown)


cinnamon-bun:
I appreciate how this communicates carefully about privacy!

cinnamon-bun:
P.S. I merged this without testing it - I don't know how to set this value myself.  Has anyone verified it works on feeds which have changed `publicWebHosting` back and forth several times?

christianbundy:
Thanks for merging this! I'll deploy it to the demo now.

> I appreciate how this communicates carefully about privacy!

:rocket: I'd love any improvements to the text, btw, I'm not crazy about it but wanted to get *something* written.

>  Has anyone verified it works on feeds which have changed `publicWebHosting` back and forth several times?

I'll check this and report back. I'd be very surprised if it were prone to a problem like #110.

Issue #177, opened 2020-02-07
Own public threads not visible in "Latest"


georgeowell:
**What's the problem you want solved?**
It feels weird to not see stuff you just posted in the chronological "Latest" feed. 

**Is there a solution you'd like to recommend?**
Own publish threads should be visible in "Latest". We will need update the feed description also. 

Relates to #176. Probably not need if latest mock ups in https://github.com/fraction/oasis/issues/170 is implemented.

Issue #155, opened 2020-02-04
Add documenation for socialFilter()


christianbundy:
Problem:

>A comment could be nice here to describe that `null` means don't apply a filter, `true`/`false` mean apply a filter for that value respectively?
>
>Also if omitted when calling the function they get these as default values?  So the function doesn't need to deal with them being undefined?
>
>(I think this makes sense, I'm just not familiar with unpacking an object and setting defaults in the function signature like this.  I don't know what the `= {}` is doing on 279)

_Originally posted by @cinnamon-bun in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0MzcwMDc3NTI5/pull_requests/unread_timeline_

Thank you, this is super useful feedback. We should add a comment here to describe how the code works. Rough summary: the `= {}` mean "if there are no arguments, default to `{}`", and the bit to the left of that says "if these properties don't exist, use these defaults", which are all used if we're passed an empty object or `undefined`. The filter works by ensuring that the `following`, `blocking`, or `me` variables are set to either `true` or `false`, and if it's `null` then it accepts either.

christianbundy:
Also I think we should probably re-enable `me` on the public pages. I was on the fence about this but after like 20 minutes I can confirm it feels uncanny to post stuff and have it show up nowhere.

PR #182, opened 2020-02-08
Fix and document `socialFilter()`

christianbundy:
Problem: The `socialFilter()` function wasn't documented and contained a
bug where it wouldn't show your posts when `following = true`. This is
because you usually don't follow yourself, so `following = true` was
basically equivalent to `me = false`.

Solution: Add some documentation and resolve the bug by adding special
handling for when the message is from us *before* passing to the general
implementation for follow/block checking.

Resolves https://github.com/fraction/oasis/issues/155
Resolves https://github.com/fraction/oasis/issues/177

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -270,6 +270,14 @@ module.exports = cooler => {
     });
   };
 
+  /**
+   * Returns a function that filters messages based on who published the message.
+   *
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
Makes sense!  Thanks for fixing this

Issue #183, opened 2020-02-09
Replace ssb-about / ssb-social-index with ssb-backlinks


christianbundy:
**What's the problem you want solved?** The ssb-about plugin is a thin wrapper around ssb-social-index, which is a wrapper around ssb-backlinks that doesn't really give us what we want. We [only care about the latest value from an author](https://github.com/ssbc/ssb-social-index/issues/2) and thanks to @cinnamon-bun's testing suggestion I just learned that [you can never change from `true` to `false`](https://github.com/ssbc/ssb-social-index/issues/3). :fire: :fire: :fire: :fire: :fire: 

**Is there a solution you'd like to recommend?** Probably just use ssb-backlinks directly.

PR #184, opened 2020-02-09
Use custom getAbout() to replace SSB-About

christianbundy:
Problem: The SSB-About plugin is incompatible with our needs. More info
in the GitHub issue linked below and in the code comments.

Solution: Unfortunately, roll our own alternative to the SSB-About
plugin so that we can be 100% sure that it pulls the latest 'about'
published by an author about themselves and doesn't just skip `false`.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -26,35 +26,81 @@ const configure = (...customOptions) =>
 module.exports = ({ cooler, isPublic }) => {
   const models = {};
 
+  /**
+   * The SSB-About plugin is a thin wrapper around the SSB-Social-Index plugin.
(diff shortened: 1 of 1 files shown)


christianbundy:
Tested this back and forth a bunch of times locally and on the oasis-demo viewer.

Resolves #183

cinnamon-bun:
Thank you so much for fixing this!

I guess the official about plug-in was also meant to be used for nicknames? and that's why it allows messages about people besides the author?

christianbundy:
Yep, I think that's the case. I think that might be useful if I wanted to give you a nickname and have it reflect locally, but I think it opens too many bugs and vectors for abuse to justify using the previous implementation as-is.

PR #186, opened 2020-02-09
2.13.0

christianbundy:
**What's the problem you solved?** We have lots of stuff that hasn't been released!

**What solution are you recommending?** Release as a new minor version!


christianbundy:
Thank you! Just pushed my tags and ran `npm publish`.

PR #189, opened 2020-02-10
2.13.1

christianbundy:
Sorry @cinnamon-bun, this is starting to feel silly. Wanted to get this patch out asap since 2.13.0 doesn't have #188

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@fraction/oasis",
-  "version": "2.13.0",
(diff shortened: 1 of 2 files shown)


cinnamon-bun:
No worries :)

Maybe releases can be exempt from don't-merge-your-own-thing.  Making releases feels more like a Maintainer action than a Contributor action

christianbundy:
That's an unreasonably good point, thank you. I'll just do that in the future.

Issue #187, opened 2020-02-09
Is it worth publishing on npm?


christianbundy:
**What's the problem you want solved?**

Publishing on npm is a total hassle and I'm not sure how worthwhile it is.

- Maintainers need to make accounts on npm, which is an extra step that provides questionable value.
- Keeping GitHub maintainers synced with npm maintainers is a chore.
- Since the `oasis` name was taken I used a `@namespace`, which apparently means you can't add maintainers to a module. You must add them to an organization, and then add them to a team, and add the team to the module. This will only take a few minutes but I imagine that means everyone has to join the 'organization' and accept the team invite? :man_shrugging: 
- If we publish on npm then now we have an `npm publish` command that we have to run just to replicate what we've already just released with `git push`.

**Is there a solution you'd like to recommend?**

We're already pushing our code and tags to GitHub, and it's trivial to have npm install the latest semver tag from GitHub:

```shell
npm --global install "fraction/oasis#semver:*"
```

Should we just do that and avoid publishing on npm? Or is it friendlier to just follow the 'publish on npm' convention and don't sweat the slightly increased complexity?

cinnamon-bun:
It'd be nice for our code to exist somewhere besides just GitHub.  That could be npm or another git hosting service.

Is that `npm` command line hardcoded to github or can it use other git hosts?

If someone wanted to use Oasis as a dependency, how would that work without npm?  For example if someone made an Oasis launcher app like in https://github.com/fraction/oasis/issues/172#issuecomment-583136508

I wish we could use [entropic](https://github.com/entropic-dev/entropic) instead of npm but it looks stalled.  No commits in 2020

christianbundy:
> Is that npm command line hardcoded to github or can it use other git hosts?

You can use any Git host, which is pretty cool IMO. I think you just have to use a full URL instead of the GitHub short name.

> If someone wanted to use Oasis as a dependency, how would that work without npm?

You can use Git URLs as dependencies, and honestly my anti-registry sentiments are leaning toward doing this for all dependencies, but it's unconventional and I really don't want Oasis to just be a bunch of experiments that are impossible to reason about. If you run ` npm install fraction/oasis#semver:^2.13.0` it should work exactly like if you ran `npm install @fraction/oasis@^2.13.0`

> I wish we could use entropic instead of npm but it looks stalled.

[...]

PR #190, opened 2020-02-10
tag and publish on npm

jedahan:
this is just a sketch, and we need to set NPM_AUTH_TOKEN

Its hard to publish on npm

If you change the release number in package.json, and make that commit message "Release x.y.z", this should tag that commit as vx.y.x . I have not tested this at all!

### DIFF
--- a/.github/workflows/npmpublish.yml
+++ b/.github/workflows/npmpublish.yml
@@ -1,0 +1,26 @@
+name: tag and publish on commit with text "Release x.y.z"
+
+on:
+  push:
+    branches:
+      - master # Change this to your default branch
+jobs:
+  npm-publish:
+    name: npm-publish
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


jedahan:
Related to #187

christianbundy:
I think I'd feel better about this if we talked a bit more about whether this is a problem first. The `npm publish` part isn't the part I'm struggling with in #187, it's more about the power balance with npm inc and whether we benefit at all from that relationship. Like, if publishing on the npm wasn't free, what would it be worth to us? The more I think about it, I keep thinking that it provides literally $0 value, and that really all we're doing is ["**contributing** packages to the registry"](https://docs.npmjs.com/packages-and-modules/contributing-packages-to-the-registry) and donating code to a for-profit company.

Could we discuss more in #187? I'm afraid that I'm not sure if this solution helps since we haven't really agreed on the problem yet (or whether it's even a problem).

PR #193, opened 2020-02-10
add screenshot to readme, fixes #59

### LINKED ISSUE #59: Screenshot in readme
**What's the problem you want to solved?** I always really enjoy seeing how software looks before I install it. I was thinking that the web demo would be good enough, but @cblgh asked for a screenshot of Oasis in the readme so that's probably a good idea.

**Is there a solution you'd like to recommend?** Maybe one of us should take a screenshot and put it in the readme? We should probably respect `publicWebHosting` and be sure not to capture any information that the author hasn't said they're okay with sharing. Bonus points for a [slow] animated GIF of some of the themes?

jedahan:
No screenshot in readme

Screenshot of demo in readme


christianbundy:
LGTM, thanks!

PR #195, opened 2020-02-11
Replace screenshot with wider screenshot of art

christianbundy:
Problem: I didn't really like random text posts from me being in the
readme, and the aspect ratio of the screenshot meant that it ended up
taking up lots of vertical space.

Solution: Use a wide screenshot of some concept art to keep it from
stretching vertically very far while still showing off the absolute
basics of Oasis.


christianbundy:
Thank you!

PR #196, opened 2020-02-11
Replace cooler methods with SSB-Client promises

christianbundy:
Problem: We started using `cooler.get()` and `cooler.read()` because it
was impossible to use promises with SSB-Client.

Solution: I made some downstream pull requests into the MuxRPC module
and the SSB-Client module, which means that both of them now natively
support promises. This commit removes the weird convenience methods and
replaces them with the native promise support, which should hopefully
make the code easier to read and write.


cinnamon-bun:
This is so much nicer!  And it will be easier to add types later, if I ever get around to that

PR #198, opened 2020-02-11
Improve readme with more details about Oasis

christianbundy:
Problem: The previous readme didn't really have very much information
about what Oasis is or what we want it to be. It wasn't very enticing.

Solution: Zoom out a bit and give a top-3 list of features that
distinguishes Oasis from other social networking applications and SSB
applications in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,28 @@
 # Oasis
 
-**Friendly Scuttlebutt interface designed for simplicity and accessibility.**
-This is an experimental client built with HTML, CSS, and Node.js without any
-front-end JavaScript. The goal is to support basic social messaging schemas with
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
Hooray for docs!  🌟 🌟 🌟 

Could be useful to link to https://scuttlebutt.nz/ ?

Also probably a good idea to mention that this is still in-progress and is missing a few core features, like joining a pub, so people will need to get started with another client before using this one

christianbundy:
Yep, I'd merge that! 

> Also probably a good idea to mention that this is still in-progress and is missing a few core features

True!

> like joining a pub

I think that was resolved by #154! :rocket:

PR #200, opened 2020-02-12
Fix sources bug and thread ancestor bug

christianbundy:
It's a two-fer. I'd do these individually but I think they should go together.

---

Problem: In the old `cooler.read()` implementation, sources returned
promises, but that isn't the case in the native SSB-Client promise
impelementation. This means that in at least one place, there's a bug
where it tries to call `.then()` on a non-promise.

Solution: Remove promise-based code from streams, which don't require
any special handling anymore.

---

Problem: Any error thrown while looking for thread ancestors was
throwing a "message not found" error, which was incorrect and useless.
This error caused me to his refresh repeatedly, not understanding that I
was publishing multiple messages. Super bad. While investigating this I
found that there's a slightly different problem where someone can post
an *invalid* message link, which we don't currently have handling for.

Solution: Only show that error when it's actually happening, and add
support for just ignoring when we see an invalid message link as `root`
or `fork`.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -7,6 +7,7 @@ const prettyMs = require("pretty-ms");
 const pullParallelMap = require("pull-paramap");
 const pull = require("pull-stream");
 const pullSort = require("pull-sort");
+const ssbRef = require("ssb-ref");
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Thanks for merging this!

PR #206, opened 2020-02-12
Fix popular page to only show posts from following

### LINKED ISSUE #202: Can see blocked content and replies to blocked users
I can currently see content from users I have blocked. I can also view their content on their profiles including blobs.

I can also see replies to threads that the people I have blocked have posted.

- Make it impossible to see content from blocked users
- Ensure replication of blocked users isn't happening
- Hide replies to blocked users

### Nice to have
- Delete content of blocked users from your local database

cc: #93

christianbundy:
Problem: Previously we were only counting likes from people you follow,
but showing messages from anyone. This was backward, and could
potentially show messages from blocked authors that were liked by people
you follow.

Solution: Move the `socialFilter()` invocation down the pipeline so that
it sorts the output messages, not the likes.

cc: @georgeowell could you verify that this resolves #202

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -836,7 +836,6 @@ module.exports = ({ cooler, isPublic }) => {
       const messages = await new Promise((resolve, reject) => {
         pull(
           source,
-          followingFilter,
           pull.filter(msg => {
             return (
               msg.value.timestamp > earliest &&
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
After 5 minutes I think I miss the old behavior where Popular showed you things from your neighbors, not just your follows.  It was a good way to find new people to follow and things going on.

When I'm reading my follows I'm more likely to read everything, not just the popular posts, because the volume is lower and I care more about them.

Gah, so many use cases!  Maybe this will be resolved in the future by configurable views.

Big plus on hiding blocked people though, that's #1 importance.

PR #207, opened 2020-02-12
Remove automatic mentions from private threads

christianbundy:
Problem: When replying to a private thread we already know who the
recipients are, and they're already going to get the notification, so
there's really no reason to add a mention.

Solution: Remove the mention when the message is private.


cinnamon-bun:
⭐️

PR #209, opened 2020-02-12
Remove roadmap now that they're all issues

christianbundy:
Problem: The roadmap should exist in the issue tracker, not the repo.

Solution: Since all of the roadmap items have been converted into
issues, we're now good to remove the roadmap from the repo.


christianbundy:
Thanks for the merge!

PR #210, opened 2020-02-12
Use color emoji in Markdown posts

christianbundy:
Problem: Emoji from Markdown posts were being displayed in the default
font, which doesn't always use color emoji.

Solution: Use the `.emoji` CSS class on those emoji to ensure that the
correct emoji font is used.


cinnamon-bun:
🐧 💯 🍀 🌾 👞

PR #213, opened 2020-02-14
[WIP] raw json submission form

black-puppydog:
This is my first attempt. Comments welcome but I'm off the train now so I'll be back on... tomorrow or on monday, :P

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -52,6 +52,7 @@ const {
   extendedView,
   latestView,
   likesView,
+  rawJsonView,
   listView,
   markdownView,
   mentionsView,
@@ -229,6 +230,9 @@ router
       });
     };
     ctx.body = await profile();
+  })
+  .get("/raw_json/", async ctx => {
+    ctx.body = await rawJsonView();
   })
   .get("/json/:message", async ctx => {
     if (config.public) {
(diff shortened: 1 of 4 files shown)


christianbundy:
Thanks for this patch! As we discussed on SSB, the problem is that Oasis doesn't let you publish all sorts of messages yet (e.g. `publicWebHosting`) and the solution would be to add a textarea to do it.

black-puppydog:
Yeah sorry I was already the last passenger on the train when everyone had left already. :D

black-puppydog:
Waaaait... Was this okay? I didn't do any error handling upon invalid json or such!

christianbundy:
Yeah! You found a problem and applied a fix. 'Good' is better than 'perfect', and it gives other people the opportunity to submit patches (like https://github.com/fraction/oasis/pull/214) for the improvements they want to make instead of nitpicking your code. Your patch was great, thank you!

If you look at the contributing guidelines you'll see that we're using a fork of the ZeroMQ development process, C4, which the original author wrote a bit more about [here](https://hintjens.gitbooks.io/social-architecture/chapter4.html), especially the bit on the [development process](https://hintjens.gitbooks.io/social-architecture/chapter4.html). I'd highly recommend checking it out if you're curious about this development style. :heart:

PR #215, opened 2020-02-14
No more `null` for unknown names and descriptions

christianbundy:
Problem: When we couldn't figure out someone's name or description we
were using `null`, which got coerced to a string and ended up with us
calling people `"@null"`. Not ideal!

Solution: Support a use-case where someone doesn't have a profile
description or a name, and use the first 8 characters of their public
key for their name if they haven't set a name themselves.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -89,10 +89,12 @@ module.exports = ({ cooler, isPublic }) => {
         return "Redacted";
       }
 
-      return getAbout({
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
⭐️

christianbundy:
Yay, thanks for merging these! You rock. :star2:

PR #218, opened 2020-02-16
Fix likes not working with Patchwork as server

christianbundy:
Problem: Recently we refactored the code so that we calls over MuxRPC
used promises, which was incompatible with the client-side SSB-Tangle
plugin that was expecting a callback. While debugging this error, I also
noticed that our `{ ws: { http: false } }` code wasn't doing anything,
and that our connection error handler was handling all kinds of
unrelated errors that would make debug more difficult.

Solution: Use `util.promisify()` to convert the callback-style function
into a promise-style function, remove the unused SSB-WS code, and reduce
the scope of the error catcher so that it'll continue to output errors
unless they're "cannot connect to sbot".


ciex:
This fix works for me!

christianbundy:
Thanks @ciex and @jedahan!

Issue #170, opened 2020-02-06
Hierarchy in sidebar to handle many navigation items


cinnamon-bun:
**Problem**

The sidebar is getting long and overwhelming.  And we're not done adding things to it yet :)  Eventually we'll winnow out pages we don't use often, but until then it will be long.

**Solution?**

I added more things to the navbar in this mockup to simulate what it'll look as we keep adding things.

Group the navbar into sub-lists:

![Screen Shot 2020-02-05 at 5 36 22 PM](https://user-images.githubusercontent.com/32660718/73898412-1d010600-483e-11ea-85cb-23cb6e46df79.png)

On narrow screens, they would do this:

![Screen Shot 2020-02-05 at 5 36 29 PM](https://user-images.githubusercontent.com/32660718/73898416-1ecac980-483e-11ea-93ed-b3c9e9cf2e6e.png)

Or if we had a navigation drawer hamburger menu on narrow screens, we could keep the first layout (the tall one).

christianbundy:
+1 on the problem that the sidebar is super overwhelming. I've been thinking that we should just have a few pages with lots of filtering options (topic/comment/any, mutuals/following/extended/both, sort by latest/popular), but that might not be a good solution. Not sure what the right solution would be here. :man_shrugging:

georgeowell:
I really like these mock-ups. I think the first layout is fine tbh. Obviously the smaller screen one would need some more thought. 

I think it's just awesome that we have (or will sometime have) so many features. You see Twitter and Facebook keep trying to "Simplify" their interfaces and it rarely looks better. It just ends up looking like kids toy software.

You could always make them collapsible with little arrows as with content warning. For example, some people would rarely open "Etc" or even "Me" if they just read stuff on SSB.

cinnamon-bun:
@christianbundy 
I like adding view options and having fewer views.  Otherwise we'll end up with every possible combination as its own view.

This one thing could replace almost every other page:

![Screen Shot 2020-02-06 at 1 42 48 PM](https://user-images.githubusercontent.com/32660718/73981209-a10ec880-48e6-11ea-94e3-83101dc61712.png)

Maybe the sidebar would be a list of editable view presets.  Oasis would ship with some defaults but you could edit/rename/delete them and add your own.

cinnamon-bun:
It'll get more complicated if we have thread-centric views.  Threads started by ___, threads including ___, threads mentioning ___, ...

georgeowell:
[...]

Issue #126, opened 2020-01-31
Show recipients of private messages


cinnamon-bun:
**What's the problem you want solved?**
In a private message you can't see who the recipients are, so you can't know the audience you're talking to.

**Is there a solution you'd like to recommend?**
Something like Patchwork, which shows recipient thumbnails in the top right corner of each message?

georgeowell:
I was going to even file an issue suggesting that we redesign the private messages interface entirely. You can probably guess where I'm going with this....

What if it looked really similar to Signal, Facebook messenger, Telegram etc? With that now standard interface of list of threads down the left hand side and thread content on the right.

cinnamon-bun:
Yeah let's try that!

To figure out:
* [ ] Layout on mobile.  Is it possible to make a hamburger-menu drawer using pure CSS?  Maybe using a restyled HTML checkbox as the toggle button
* [ ] If you have multiple threads with the same person would they get combined into a single sidebar item?

I've actually wanted to try kind of layout even for public threads.  Like Thunderbird, a list of threads on the left and the messages in a thread on the right.  I think it might help you keep track of all the threads that are always in play at the same time.

georgeowell:
Manyverse has private messages now and it looks very chat like.

cbleslie:
@christianbundy Show Stopper

Powersource:
uuh how do i assign @lindskog to this issue

christianbundy:
plz don't unassign me

christianbundy:
> uuh how do i assign @lindskog to this issue

GitHub only lets you assign collaborators (trusted folks who have contributed to the repository)

Powersource:
I wanted to assign @lindskog to make it clear to the other new workers that the issue is being worked on.

christianbundy:
Issues can have multiple assignees though.

> make it clear to the other new workers that the issue is being worked on

I'd just comment that. If you *need* this to be searchable with the GitHub UI you can make a custom "powersource knows someone working on this" label or something. Or organize your work with a spreadsheet or something? I'm not sure.

christianbundy:
Just switched my notifications to PRs only, please @mention me if you need my attention in issues. :heart:

cbleslie:
@christianbundy are we there yet? I want a drink of water. @Powersource keeps moving onto my side of the car! /whine

lindskog:
PR for this issue: https://github.com/fraction/oasis/pull/584

PR #219, opened 2020-02-16
Show a preview of comments in listings

ciex:
First of all, I am not familiar with the C4 contribution process, please be patient with me :) I just built this for myself today because I wanted to get familiar with the Oasis codebase and wanted to have this feature myself. If it's not in your vision for Oasis, no problem.

On a large monitor, Oasis is only using a fraction of the horizontal space. Also, I was missing Patchworks feature to show a preview of comments on each post in the timeline.

Example:

image

Here I have added a preview of comments on each post in the timeline, which is shown to the side of the original post on large screens and below on small screens (as in Patchwork).

Examples:

image

image

**Caveats**

- The popular page now takes a long time to load (56sec vs 12sec before). Both times are too long to be usable though. I wonder what would be a good way to speed up loading of the page but I feel like I don't know enough about the architecture and also that this would be out of scope for this maybe (?).
- I duplicated a bit of code here that I would like to refactor before this is mergeable. I wanted to hear whether this approach is even wanted before doing the work though. Hope you understand!
- I had a problem committing my changes when Git's commit hook would ask me to run prettier on the code first. As there is no prettier config in the project, that would change all of the rest of the codebase, which is probably not what is intended. So I suggest to add a `.prettierrc.js`.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -331,6 +331,7 @@ section {
   word-wrap: break-word;
   background: var(--bg);
   width: 100%;
+  max-width: var(--measure);
   box-sizing: border-box;
 }
 
@@ -396,6 +397,32 @@ section > footer > * {
 
 section > footer > form > button.liked {
   color: var(--red);
+}
+
+.post-aside-wrapper {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+}
+
+@media (min-width: 80rem) {
+  .post-aside-wrapper {
+    flex-direction: row;
+  }
+}
+
+.post-aside  {
+  display: flex;
+  flex-wrap: wrap;
+  flex-direction: column;
+  align-items: flex-start;
+  padding-left: var(--micro);
+  margin: var(--whole) 0;
+}
+
+.post-aside section  {
+  max-width: var(--peta);
+  margin: 0 0 var(--micro) 0;
 }
 
 label {
@@ -495,8 +522,7 @@ hr {
     margin-bottom: var(--whole);
   }
   main {
-    width: 100%;
-    max-width: var(--measure);
+    width: 85vw;
   }
   body {
     display: flex;
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -411,7 +411,7 @@ section > footer > form > button.liked {
   }
 }
 
-.post-aside  {
+.post-aside {
   display: flex;
   flex-wrap: wrap;
   flex-direction: column;
@@ -420,7 +420,7 @@ section > footer > form > button.liked {
   margin: var(--whole) 0;
 }
 
-.post-aside section  {
+.post-aside section {
   max-width: var(--peta);
   margin: 0 0 var(--micro) 0;
 }
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -129,7 +129,7 @@ const postInAside = msg => {
     key: encodeURIComponent(msg.key),
     author: encodeURIComponent(msg.value.author),
     parent: encodeURIComponent(msg.value.content.root)
-  }
+  };
 
   const url = {
     author: `/author/${encoded.author}`,
@@ -140,45 +140,46 @@ const postInAside = msg => {
     json: `/json/${encoded.key}`,
     reply: `/reply/${encoded.key}`,
     comment: `/comment/${encoded.key}`
-  }
+  };
 
-  const isPrivate = Boolean(msg.value.meta.private)
-  const isRoot = msg.value.content.root == null
-  const isFork = msg.value.meta.postType === "reply"
-  const hasContentWarning = typeof msg.value.content.contentWarning === "string"
+  const isPrivate = Boolean(msg.value.meta.private);
+  const isRoot = msg.value.content.root == null;
+  const isFork = msg.value.meta.postType === "reply";
+  const hasContentWarning =
+    typeof msg.value.content.contentWarning === "string";
   const isThreadTarget = Boolean(
     lodash.get(msg, "value.meta.thread.target", false)
-  )
+  );
 
   // TODO: I think this is actually true for both replies and comments.
-  const isReply = Boolean(lodash.get(msg, "value.meta.thread.reply", false))
+  const isReply = Boolean(lodash.get(msg, "value.meta.thread.reply", false));
 
-  const timeAgo = msg.value.meta.timestamp.received.since.replace("~", "")
+  const timeAgo = msg.value.meta.timestamp.received.since.replace("~", "");
 
   const markdownContent = markdown(
     msg.value.content.text,
     msg.value.content.mentions
-  )
+  );
 
   const likeButton = msg.value.meta.voted
     ? { value: 0, class: "liked" }
-    : { value: 1, class: null }
+    : { value: 1, class: null };
 
-  const likeCount = msg.value.meta.votes.length
+  const likeCount = msg.value.meta.votes.length;
 
-  const messageClasses = []
+  const messageClasses = [];
 
   if (isPrivate) {
-    messageClasses.push("private")
+    messageClasses.push("private");
   }
 
   if (isThreadTarget) {
-    messageClasses.push("thread-target")
+    messageClasses.push("thread-target");
   }
 
   if (isReply) {
     // True for comments too, I think
-    messageClasses.push("reply")
+    messageClasses.push("reply");
   }
 
   const postOptions = {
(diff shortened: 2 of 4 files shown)


src/assets/style.css
```diff
@@ -495,8 +522,7 @@ hr {
     margin-bottom: var(--whole);
   }
   main {
-    width: 100%;
-    max-width: var(--measure);
+    width: 85vw;
```
ciex: Need to change this to something that doesn't conflict with the nav bar on small screens

src/views/index.js
```diff
@@ -56,6 +56,8 @@ exports.setLanguage = language => {
 const markdownUrl = "https://commonmark.org/help/";
 const doctypeString = "<!DOCTYPE html>";
 
+const THREAD_PREVIEW_LENGTH = 3;
```
ciex: I didn't see a configuration file for constants like this so I added it inside the module

src/views/index.js
```diff
@@ -121,6 +123,188 @@ const template = (...elements) => {
 
   return result;
 };
+
+const postInAside = msg => {
```
ciex: so this is duplicating a lot of code from `post()`. My approach here would be to add a library with utilites for working with post objects to solve things like assembling the object of URL's for a post.

christianbundy: This would be cool! I've been thinking that maybe the router (`src/index.js`) should be handing these URL builders, and I think koa-router even supports this.

christianbundy:
Hey, thanks for this patch! It's really cool to see you diving into the code and making changes. :rocket: I think it might make sense to spend a bit more time discussing the problems to make sure we agree on what the problem(s) are and how we want to solve them. As I'm sure you've seen, the code is sort of bleh and I know there are some changes that will touch lots of the same code that you touched.

Right now we're in this weird transitional page where we have a page for every type of view, whereas we should probably have filter/view options [like this](https://github.com/fraction/oasis/issues/170#issuecomment-583126856). I think it'd be really really cool to have a thread preview, or "thread rollup" as it's called in Patchwork. I know it isn't ideal, but maybe we could experiment with this by making a new page called "Summaries" or something? I'm imagining it having the same data as "Topics", but we could do a different page if you preferred. Maybe others would have a good idea on how to do this best? I'd love if you could open an issue about summaries so we could all discuss and figure out what we want in a solution. I imagine it'd look a lot like what you've built. :heart: For example, you might have already written some of the code that would resolve #220!

Regarding the thread redesign to target wide screens, what happens when a comment is longer than the original post? I'd love to get feedback from others on this solution, since it isn't something I've seen on other social media, but maybe my initial reaction would change after playing around with it a bit more. I think I'd be really happy with this sort of design if we had an email-style interface where the left side was a summary and the right side was the thread, as discussed [here](https://github.com/fraction/oasis/issues/126#issuecomment-581035222), but I think it's something I'd like to discuss more first to make sure we agree on a problem and have a solution that fits.

> I had a problem committing my changes when Git's commit hook would ask me to run prettier on the code first.

`npm run fix` should solve this, but it should be documented! Or we should use lint-staged or something so this happens automatically...

jedahan:
I really like this experimentation, and take it as a good example that oasis-as-a-platform is in a healthy place to encourage new ideas.

I wonder how we feel about making it easier to incorporate more radical experiments within oasis itself. Like an experiments or labs setting to allow for people to try different things. Just my initial thoughts.

jedahan:
And more specifically on this two-dimensional design, I think its a real cool way of taking advantage of real estate in a novel manner, and differentiating from the monotony of 'getting through a single infinite feed', encouraging looking at the conversation a bit more deeply and with good context.

cinnamon-bun:
Thanks for this experiment!  It's a promising idea - I miss Patchwork's way of showing threads.

❗️ Rambling thoughts ahead ❗️ 

I've been trying to figure out how all these different view styles will encourage or discourage healthy community, and I'm feeling stumped.

### The kind of interaction I'd like to encourage

The feeling I like best is long conversation threads where each message is a paragraph or more and the conversation goes back and forth for a long time.  People engaging with each other deeply, like pen pals.  This happens sometimes on Hacker News though it's more idea-based instead of person-based; on Metafilter; and in Patchwork.  Livejournal also used to feel this way at times.  Github issue conversations are another good example.

Anti-examples are Twitter and Mastodon (too short and scattered), Reddit and Youtube (full of low-effort memes), and Facebook (sometimes long posts, but short replies that don't build on each other).

Oasis has a Twitter-like feel now, where each comment is scattered around instead of seen in context, and I find it confusing.  I prefer Patchwork's thread-centric views.  It makes you re-read the last couple of messages before replying, which I think encourages long thoughtful conversations that build on themselves.

I like the idea of treating **initial posts and comments as equally important**, to encourage these long conversations.  (As opposed to, say, Youtube comments, where each comment is a standalone response to the video without much interaction between comments).  So more of a chat-conversation-thread model instead of a post-and-short-comments model.

### Warrens and Plazas

Another dimension is **warren-vs-plaza**.  Warrens hide conversations in side rooms you have to click into, and plazas mix everything into one more public feeling space.

I'd predict that **warrens** feel more intimate and would encourage better conversations.  Examples: email threads, slack channels, chat groups in Signal.  This also helps confine annoying topics in their own warrens (like tech talk about SSB).  The downside is the more structured interface makes you click around to see into each warren, and posts might evolve clickbait subject lines to attract attention.

Patchwork is more of a **plaza** and I like it more than I thought I would.  It shows you the richness of conversations happening all around, and you can step in and join them when they become interesting.  It helps conversations gather more replies and last longer.  But it can become an overwhelming firehose when everyone's posting at once.

### Sun mode and Moon mode

In conclusion, imagine something like Github issues with dual ways of approaching the content: a structured list of threads, and a way to overhear the latest activity across all threads.

Mix and Zach have vocabulary for this:
%VzxV0Je3iaLmWEVCG2KqyvIZzFRwwfFTqXZLeD5at5g=.sha256

"☀️ Sun mode" and "🌘 Moon mode".  Sun mode is linear, purposeful, and directed.  Moon mode is coincidental, organic, wandering.  Maybe we could have both?

ciex:
Thank you all so much for your thoughts <3 I am glad that some of you like the idea of branching out the timeline horizontally. I know this is quite experimental but I have been using it yesterday and today and somehow it just feels right to mee (except for some edge cases). I would post a video with some more scrolling here but I feel weird about putting people's scuttling on Github - so I encourage you to give it a try!

Regarding those edge cases: @christianbundy you asked about the layout if the comments end up requiring more height than the original post. My own intuition is that the "wrap-up" or aside should be limited in height to be able to fit the visible viewport height on a common 13 inch display if the original post is shorter than that. If the original post is longer than that the aside should not continue flowing past the original post.  We could either use Javascript to look at the rendered height of containers and react to that or do something by counting the length of the post's text content and ignore image height.

@cinnamon-bun I really like the ux perspective you are taking. I would add to that this critique of timelines for social media `%254UHRF%2FxbCaeyk%2FJgw92LBWXrfGm9SCo3YIfAVXwtNzw%3D.sha256`. (Side note: How can I make Oasis show me a post by giving it a message key? Putting it in search doesn't work for me so I just edited the URL, but that shouldn't be the way should it?)

Another experiment I would like to do is to create another kind of "popular" view that allows users to reflect on partciular time period (e.g. what was Scuttlebutt about in Winter 2019), I think I will write about that on ssb and link it here later.

I created a discussion issue about summaries (#221) as suggested by @christianbundy .

**Regarding this PR**: What is the process here? Should it wait for that discussion? Who decides whether this can go forward? I would love to polish it up if you can signal me that it would be merged eventually. I could also put in a config option to make it optional.

ciex:
Oh by the way: Everything is way faster when Patchwork is running in the background.

jedahan:
Generally our dev process has been: if it doesn't obviously *break* core features, we merge optimistically and then make improvements.

In that spirit the only thing I *might* consider broken is the 5x slowdown for popular. It would be a shame to block merging on just that, so either a global option, or turning it off just on the popular page would be enough for me to hit the big green merge button. Gonna try it out for the rest of the day.

christianbundy:
> Generally our dev process has been: if it doesn't obviously *break* core features, we merge optimistically and then make improvements.

Right, specifically when the patch is "a minimal and accurate answer to exactly one identified and agreed problem". I'm going to optimistically merge this, let's see where it goes! Thanks for this patch @ciex.

ciex:
Oh, now that was unexpected (ans also pretty cool). Actually, I would have liked to refactor some of things that I already commented on this PR. But I guess then I'll just make another PR for that.

PR #223, opened 2020-02-17
Create 'Summaries' page for new summary view

christianbundy:
Problem: Until we have a way to change the view of each page, it seems
to me that each of the views that we're experimenting with should be on
their own page rather than modifying the existing pages to add more
functionality. I'm also noticing that I can't process all of the
information that's on the screen when the new horizontal layout is used,
and I think we would benefit from using the previous layout
consistently.

Solution: Move the summary view to its own page so that we can continue
to iterate on it without modifying the popular view, which is already
complex, and revert the style changes to maintain consistency with the
rest of the linear design (threads go downward, indentation denotes
sub-threads, etc).

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -331,7 +331,6 @@ section {
   word-wrap: break-word;
   background: var(--bg);
   width: 100%;
-  max-width: var(--measure);
   box-sizing: border-box;
 }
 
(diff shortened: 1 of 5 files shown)


georgeowell:
I agree! It's fine to be experimental at this stage but we don't want to disrupt the work we've already done.

georgeowell:
I was quite surprised when I opened Oasis and everything was aligned to the left :scream:

christianbundy:
Thanks for the merge!

cinnamon-bun:
Thanks @christianbundy !  Also thanks again to @ciex for the initial patch.

As a side note I was wishing the thread preview showed the *last* 3 comments instead of the *first* 3.  This would match Patchwork and make it easier to keep up with new comments as they appear.

christianbundy:
@cinnamon-bun I *think* you could adjust that by changing `slice()` here: https://github.com/fraction/oasis/blob/16f19e85fbff2b656dd0f9f361bb3ae9f9da4e20/src/views/index.js#L299-L307

PR #225, opened 2020-02-17
Redirect search to SSB references

christianbundy:
Problem: There wasn't an easy way to copy and paste a message ID, feed
ID, or blob ID and get directed to that page in Oasis.

Solution: When someone pastes one of the above into the search, just
redirect the browser to the correct page rather than searching for that
text.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,6 +34,7 @@ const ssbMentions = require("ssb-mentions");
 const ssbRef = require("ssb-ref");
 const isSvg = require("is-svg");
 const { themeNames } = require("@fraction/base16-css");
+const { isFeed, isMsg, isBlob } = require("ssb-ref");
 
 const ssb = require("./ssb");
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Thank you!

cinnamon-bun:
Hooray, this makes it much easier to jump between Patchwork and Oasis :D

christianbundy:
Woooo!

P.S. Maybe the search should just be always available on the sidebar? :man_shrugging:

jedahan:
I love always available search.

cinnamon-bun:
Would it make sense to do this for hashtags also?  Searching for #foffee sends us to

http://localhost:3000/search?query=%23foffee

Which is different than clicking a hashtag link in a message, which goes to

http://localhost:3000/hashtag/foffee

I'm unclear on the differences between those two views.  `/hashtag/` is looking for SSB channels (which we don't render)?

christianbundy:
> Would it make sense to do this for hashtags also?

Yep, I think that's a great idea.

> `/hashtag/` is looking for SSB channels (which we don't render)?

Yeah, I'm not excited about supporting channels or encouraging their use unless we can really nail the ergonomics, but we should maybe at least do something like:

```javascript
if (message.value.content.channel != null) {
  message.value.content.text += `\n\n#${message.value.content.channel}`
}
```

So that if someone publishes a message with channel #foo, we at least display it at the end of their message.

#foo

Issue #49, opened 2020-01-10
Hide messages that you're done with


christianbundy:
**What's the problem you want to solved?** I have lots of trouble holding everything in my head at once and the mentions page stresses me out. Once I'm done with a message I'd like for it to be hidden somehow, but I'm not sure the best way to do this.

**Is there a solution you'd like to recommend?** Maybe each message should have a 'archive' button and each feed should have a way to filter those out? I know that Patchbay has an 'unread' plugin that marks a message as 'read' once it appears on your screen for some amount of time, but I think there are two problems:

- I read lots of stuff that I want to reply to later. I'd rather have this be an explicit thing where I choose when I'm done with a message.
- The unread plugin exists outside of the SSB log, so it doesn't persist via SSB. If we make this an explicit manual action I think that we can safely post this as a private message to ourselves, whereas making it automatic probably makes the volume of messages unreasonably large.

What's the best way to do this?

cinnamon-bun:
Email workflows tend to use 3 bits of information: read/unread, inbox/archived, and starred-for-later.

Starred-for-later would work well on both private and public messages.

Inbox/archived would let you do Inbox Zero on private messages.  Maybe less useful on the firehose of public messages.

For read/unread, I'm not sure how to know when something has been meaningfully seen.  Maybe only if you look at the thread detail view?

Stars and inbox/archived could be private messages to yourself, and read/unread could be offchain since it will be so much data.

---

For bookmarking and workflow reasons I'd love to have general purpose tagging of posts which is onchain and private-to-yourself, but that's getting out of scope :)

cinnamon-bun:
The ability to mute a thread would be really nice for safety / calm / curation reasons.

It could be built on the same infrastructure as stars, read/unread, etc.

christianbundy:
Thanks for thinking this through! I totally forgot about 'save for later', I think that would improve the "read" / "unread" thing. Dominic mentioned that he has to use "mark as unread" all the time on GitHub, which is sort of a silly and confusing semantic when *really* you just need a 'save'.

I think I need to generalize the filter on the 'popular' pages so that you can do filters on others -- even just switching between 'inbox' and 'archive' would be super sweet on the 'private' page.

> The ability to mute a thread would be really nice for safety / calm / curation reasons.

Yup! I've been meaning to implement this but haven't come up with a good UI for un-muting. Do you think we'd need a view for muted threads, or have some other way of doing it?

cinnamon-bun:
> generalize the filter

[...]

PR #230, opened 2020-02-18
Replace messagesByType() with SSB-Query read

christianbundy:
Problem: The `messagesByType()` queries don't let us sort by asserted
time, which means that we can get messages from a long time ago that
just ended up on our computer.

Solution: Use SSB-Query everywhere. The query syntax isn't really my
favorite but it lets us make declarative queries that respect the
asserted timestamp instead of just the received timestamp.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -705,12 +705,23 @@ module.exports = ({ cooler, isPublic }) => {
 
       const myFeedId = ssb.id;
 
-      const options = configure({
-        type: "post",
-        private: false
-      });
-
-      const source = ssb.messagesByType(options);
+      const source = ssb.query.read(
+        configure({
+          query: [
+            {
+              $filter: {
+                value: {
+                  content: {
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
Do we generally prefer to sort by asserted time instead of received time?

Received time is nice for bringing new-to-you items to the top, but it's pretty jumbled after a big initial sync.

Asserted time might make conversations stay more in order, but can be spoofed.

Cabal is having a similar UX challenge - what to do when old messages arrive later.  Do they go into the backscroll, or down with the new messages so they'll get noticed?

christianbundy:
I think asserted time, with the assumption that we trust that our friends aren't spoofing message times, but I'd be happy to change my opinion also. It seems like maybe mentions/private should go by received time though maybe, at least until we have #49. Maybe asserted by default, received for feeds that act like a queue of messages you should be dealing with?

PR #232, opened 2020-02-18
Add Oasis version to settings

christianbundy:
Problem: It was impossible to tell which version of Oasis you're
running, which makes it hard to know when you need to update.

Solution: Put the version in the settings so it's easy to reference.


jedahan:
Thank you!

christianbundy:
No problem, hope this helps!

PR #233, opened 2020-02-18
Redirect to Latest after publishing new message

christianbundy:
Problem: After publishing a message you were redirected to the Popular
view, which wasn't good feedback that your message had been published.

Solution: After publishing a new message, redirect to the Latest view.


cinnamon-bun:
👍  Latest loads faster too.

PR #234, opened 2020-02-18
Remove image border

christianbundy:
Problem: I added an image border to try to make it easier to see the
boundaries of images but it's not really necessary and at least one
person didn't love it, which is grounds for removal.

Solution: Get it out of here! This commit removes the image border and
also removes a random `console.log()` that shouldn't have been left in
`src/index.js`.


christianbundy:
Thanks @cinnamon-bun! :heart:

PR #235, opened 2020-02-18
Add friendly EADDRINUSE error message

christianbundy:
Problem: When the port was already taken by another process we've been
dumping a big esoteric error message that really isn't helpful for most
people.

Solution: Catch the error and give our users a friendly error message
with a suggestion on how to run Oasis on another port.


cinnamon-bun:
🌟

PR #236, opened 2020-02-18
Remove private messages from Mentions

christianbundy:
Problem: Private messages should only be showing up in the Private page,
not the Mentions page, but right now they're showing up in Mentions.

Solution: Add a check to ensure that Mentions doesn't have any private
messages.


christianbundy:
Thank you!

PR #244, opened 2020-02-19
Remove private posts from public views

christianbundy:
Problem: During the SSB-Query refactor I used `{ private: false }` in an
attempt to hide private messages, but apparently this option is only
available via SSB-Backlinks, so it was ignored without any warning (!).
This allowed private messages to be displayed in public views.

Solution: Remove the unused option and add a filter to public views that
ensures that private messages are always hidden.

cc: @georgeowell

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -20,6 +20,10 @@ const defaultOptions = {
   meta: true
 };
 
+const publicOnlyFilter = pull.filter(
+  message => lodash.get(message, "value.meta.private", false) === false
+);
+
 /** @param {object[]} customOptions */
 const configure = (...customOptions) =>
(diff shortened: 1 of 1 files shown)


christianbundy:
@georgeowell did you test this locally / can you confirm that it works? No trouble if not, just thought I'd ask just in case. Two heads are better than one!

cinnamon-bun:
Some context for the general public:

This didn't expose any private messages to people who shouldn't have seen them.  (They can't decrypt them!)

It's just that the private messages YOU could see were being shown mixed in with the public messages, on your computer only.  Not a privacy problem, just a confusing user experience.

PR #246, opened 2020-02-19
Update SSB-Mentions and remove workaround

christianbundy:
Problem: We've been using a workaround to filter the results of
SSB-Mentions, but that's just been fixed upstream and so we don't need
the workaround anymore.

Solution: Update the SSB-Mentions package and remove the workaround.

See: https://github.com/ssbc/ssb-mentions/pull/17

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1782,14 +1782,6 @@
         "regexp.prototype.flags": "^1.2.0"
       }
     },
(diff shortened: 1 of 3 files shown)


cinnamon-bun:
🏆 for digging into and fixing so many dependencies!

PR #250, opened 2020-02-21
Add support for Unix socket with noauth

christianbundy:
Problem: We're doing tons of unnecessary cryptography by encrypting the
connection between the "client" and "server", which are often running in
the same process.

Solution: Instead of connecting to the SSB service over TCP and
encrypting the stream, just connect over a socket (supported on Windows,
macOS, and Linux) and don't bother encrypting anything. This is what
Patchwork and Patchbay do already, and since our secret is at
`~/.ssb/secret` then we should be comfortable with `~/.ssb/socket` being
a trusted file where access implies authentication.

Local tests suggest that when sodium-native is available, this commit
reduces the time to render the 'Popular (Day)' page by 17%, but when we
have to fall back to JavaScript cryptography the same page now takes 30%
less time to render. My intuition is that this improvement is more
dramatic on mobile, but requires further testing before we can pat
ourselves on the back too much. :)

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -28,6 +28,7 @@
     "multiserver",
     "muxrpc",
     "nanohtml",
+    "noauth",
     "nosniff",
     "paramap",
     "patchbay",
--- a/src/ssb.js
+++ b/src/ssb.js
@@ -10,6 +10,16 @@ const ssbConfig = require("ssb-config");
 const flotilla = require("@fraction/flotilla");
 const ssbTangle = require("ssb-tangle");
 const debug = require("debug")("oasis");
+const path = require("path");
+
+const socketPath = path.join(ssbConfig.path, "socket");
+const publicInteger = ssbConfig.keys.public.replace(".ed25519", "");
+const remote = `unix:${socketPath}~noauth:${publicInteger}`;
+
+// This is unnecessary when https://github.com/ssbc/ssb-config/pull/72 is merged
+ssbConfig.connections.incoming.unix = [
+  { scope: "device", transform: "noauth" }
+];
 
 const server = flotilla(ssbConfig);
 
@@ -22,7 +32,7 @@ const log = (...args) => {
 
 const rawConnect = () =>
   new Promise((resolve, reject) => {
-    ssbClient()
+    ssbClient(null, { remote })
       .then(api => {
         if (api.tangle === undefined) {
           // HACK: SSB-Tangle isn't available in Patchwork, but we want that
@@ -64,7 +74,9 @@ const createConnection = config => {
               resolve(ssb);
             })
             .catch(e => {
-              log(e);
+              if (e.message !== "could not connect to sbot") {
+                log(e);
+              }
               connectOrRetry();
             });
         };


jedahan:
would love to have a phone conversation to talk about this so I may understand the implications, but if no one approves in the next 3 days I will test it out.

christianbundy:
Wanna talk now? I can hop on Jitsi or something to discuss, or we could schedule it at some point in the future if now is inconvenient.

christianbundy:
Brain dump for anyone interested: Right now we spin up an SSB service in the background, which is like `ssb-server start` or similar. This happens in `ssb.js`, and this service runs in the background until Oasis is closed.

When we want information from the SSB service, as we usually do in `models.js`, we connect to it over TCP with SSB-Client. The variable is called `cooler`, but it's just an SSB-Client that integrates more deeply with the SSB service so that it can auto-reconnect or auto-restart if there are problems.

The thing about the TCP connection is that *anyone* can connect, so to prove that we aren't some rando we use Secret Handshake and Box-Stream to encrypt the stream with our public key. When the server sees that the client is using the same key, the SSB-Master plugin gives it permission to run any command it wants (whereas randos can only run `createHistoryStream()` and a few others). 

The problem with this approach is that TCP, Secret Handshake, and Box-Stream are all super resource-intensive, especially on low-end devices or environments where we use JavaScript cryptography (which is the fallback when sodium-native won't compile). Ideally we want some way to authenticate the client, so that we know it's not a rando, and it would be nice if we could also avoid TCP.

The most popular solution to this problem is to layer two solutions together:

- **Unix Socket:** Instead of using TCP, we can use a Unix socket, which is a magical file where you can pipe data in one side and out the other without touching the networking stack. You can make one with `mkfifo /tmp/example-socket`, and then listen to it with `tail -f /tmp/example-socket`. In another terminal, run `echo hello > /tmp/example-socket` and you'll see your text pop up in the original terminal. This removes the networking overhead, and if we put the socket somewhere protected (like `~/.ssb/socket`) then we can be sure that it's not accessed by randos.
- **Noauth:** Since we know that we're the only people accessing this socket, we really don't need to encrypt the stream between the client and the SSB service. Anyone who can access the socket could also access `~/.ssb/secret`, so we can have a policy where we trust anyone with access to the socket. With that in mind, we really don't need to encrypt the stream with Secret Handshake or Box-Stream, so we can just turn off authentication whenever we receive a connection via the Unix socket.

Happy to answer questions, jump on a call, or both! :heart:

cinnamon-bun:
Looks good to me! Thanks for the long explanation. I'll wait to merge in case @jedahan has concerns

jedahan:
That is a fantastic explanation, and thank you for taking the time to flesh out how it all works. As long as the socket is created with restriced permissions (im guessing 600 is fine?) then merging this seems good.

The only weird place I can think of is if we are connecting to another ssb-server who has setup the socket differently, or not at all, to fallback to TCP if the socket is not under our control.

That can be an improvement for another PR though.

christianbundy:
No problem! It's helpful for me to remember how it all works when I'm trying to explain it, so double win.

> As long as the socket is created with restriced permissions (im guessing 600 is fine?) then merging this seems good.

:+1: 

```console
stat ~/.ssb/socket
  File: /home/christianbundy/.ssb/socket
  Size: 0         	Blocks: 0          IO Block: 4096   socket
Device: 802h/2050d	Inode: 1201650     Links: 1
Access: (0600/srw-------)  Uid: ( 1000/christianbundy)   Gid: ( 1001/    sudo)
Access: 2020-02-21 13:19:35.335945903 -0800
Modify: 2020-02-21 13:19:35.332612603 -0800
Change: 2020-02-21 13:19:35.332612603 -0800
 Birth: 2020-02-21 13:19:35.332612603 -0800
```

> The only weird place I can think of is if we are connecting to another ssb-server who has setup the socket differently, or not at all, to fallback to TCP if the socket is not under our control.

Yep, although as long as we're only promising "Oasis works by itself or with Patchwork" I think we're good. In the future I'd love to support other servers, but we'd also have problems with different plugins being loaded on the server.

PR #252, opened 2020-02-22
Read settings from $XDG_CONFIG_HOME/oasis/config.json

black-puppydog:
This should give pretty much expected behaviour. Each config value
can be set by three sources:

1. By command-line argument. If it is not given, then
2. By config file. Or, lastly
3. By default value in the source code.

I can't test that the config file is searched and read from the right
place on windows or macOS, but on linux it works.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,12 +4,24 @@
 
 // Koa application to provide HTTP interface.
 
+const fs = require("fs");
+const envPaths = require("env-paths");
+const config_file = envPaths("oasis", { suffix: "" }).config + "/config.json";
+console.log(`Trying to read config file ${config_file}`);
+var presets;
+if (fs.existsSync(config_file)) {
+  presets = require(config_file);
+} else {
+  presets = {};
+}
 const cli = require("./cli");
-const config = cli();
+const config = cli(presets);
 
 if (config.debug) {
   process.env.DEBUG = "oasis,oasis:*";
 }
+
+console.log(JSON.parse(JSON.stringify(config)));
 
 process.on("uncaughtException", function(err) {
   // This isn't `err.code` because TypeScript doesn't like that.
@@ -22,6 +34,8 @@ It might be another copy of Oasis or another program on your computer.
 You can run Oasis on a different port number with this option:
 
     oasis --port ${config.port + 1}
+
+Or you can set a default port in ${config_file}
 `
     );
   } else {
@@ -40,7 +54,6 @@ process.argv = [];
 const http = require("./http");
 
 const debug = require("debug")("oasis");
-const fs = require("fs").promises;
 const koaBody = require("koa-body");
 const { nav, ul, li, a } = require("hyperaxe");
 const open = require("open");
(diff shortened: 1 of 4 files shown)


black-puppydog:
hold it... swapping in env-paths for some hand-written one-liner actually seems to have broken this... :|

christianbundy:
Hey, thanks for the patch! I :heart: XDG BDS.

Could you maybe add some more info on what problem this solves?

black-puppydog:
Oh, yeah. At work I have an application that always holds port 3000. So I wanted a way to launch oasis more easily.
I'd imagine that it would also be a good way to make sure a public instance is always in public mode, and responding to the right host name. Generally, anything that can be configured by command line seems like a good candidate for file based configuration.

christianbundy:
Thanks, that's good info to have -- I can understand your use-case lots better now.

How can I go about testing this? I've put a file in `~/.config/oasis/config.json` with the contents:

```json
{
  "port": 4224
}
```

But when I run `./src/index.js --help` it still shows `3000`. I'm sure I'm doing something wrong, could you help me test this better?

src/index.js
```diff
@@ -4,13 +4,26 @@
 
 // Koa application to provide HTTP interface.
 
+const fs = require("fs");
+const envPaths = require("env-paths");
+const config_file =
+  envPaths("ssb-oasis", { suffix: "" }).config + "/config.json";
```
christianbundy: Found the reason `~/.config/oasis` wasn't working!

black-puppydog: yeah. I wanted to just use oasis, but on debian/ubuntu that's already an application that seems quite a bit older. I don't know if they use ~/.config/oasis but I didn't feel like pushing it...

black-puppydog: okay I'm back from the dead. really needed a good... 6h30 nap. :D
okay, I [searched for XDG_CONFIG_HOME in the other oasis' repo](https://github.com/ocaml/oasis/search?q=XDG_CONFIG_HOME&unscoped_q=XDG_CONFIG_HOME) and came up empty. I'll force-push the version with just ~/.config/oasis then. :)

src/index.js
```diff
@@ -4,13 +4,25 @@
...
+}
 const cli = require("./cli");
-const config = cli();
+const config = cli(presets);
 
 if (config.debug) {
   process.env.DEBUG = "oasis,oasis:*";
 }
 
+console.log(JSON.parse(JSON.stringify(config)));
```
christianbundy: I'm concerned that this is going to print private keys -- could you confirm that isn't the case before we merge? Sorry, I hate to nitpick but I'm afraid someone will copy and paste the startup output and accidentally leak their private key somewhere.

black-puppydog: valid concern, but this is the oasis config file, not the config from `~/.ssb`. Is there a mechanism that combines the two somehow?
On a code level, that line is in `index.js` and is executed way before we even `require("./ssb");`
For me, it prints this:

```js
{ _: [],
  open: false,
  offline: false,
  host: 'localhost',
  port: 13132,
  public: false,
  debug: true,
  '$0': 'oasis' }
```

I'd be happy to figure out how to get `_` and `$0` out of there, but I don't see how the ssb config would be printed here, hence no key material.

christianbundy: Oops, you're totally right. Thanks for reminding me how that worked!

PR #253, opened 2020-02-22
Refactor default config loader

christianbundy:
Problem: We load the default config with `require()`, which would allow
arbitrary JavaScript execution, are producing debug output everywhere
with `console.log()` statements, and don't have instructions for how to
write the config file. The config file is also `.config/**/config.json`,
and I think it would be helpful to specify that this is the **default**
config rather than any other kind of config.

Solution: Rename the config to `~/.config/oasis/default.json`, replace
the `require()` with `fs.readFile()`, remove `console.log()` for
debugging, and add some example JSON for how to use the config file.

cc: @black-puppydog in case you feel very strongly about `~/.config/oasis/config.json` :heart:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,22 +6,35 @@
 
 const fs = require("fs");
 const envPaths = require("env-paths");
-const config_file = envPaths("oasis", { suffix: "" }).config + "/config.json";
-console.log(`Trying to read config file ${config_file}`);
-var presets;
-if (fs.existsSync(config_file)) {
-  presets = require(config_file);
-} else {
-  presets = {};
+const path = require("path");
+
+const defaultConfigFile = path.join(
+  envPaths("oasis", { suffix: "" }).config,
+  "/default.json"
+);
+
+const defaultConfig = {};
+
+try {
+  const defaultConfigOverride = fs.readFileSync(defaultConfigFile, "utf8");
+  Object.entries(JSON.parse(defaultConfigOverride)).forEach(([key, value]) => {
+    defaultConfig[key] = value;
+  });
+} catch (e) {
+  if (e.code === "ENOENT") {
+    // No default config, no problem.
+  } else {
+    console.log(`There was a problem loading ${defaultConfigFile}`);
+    throw e;
+  }
 }
+
 const cli = require("./cli");
-const config = cli(presets);
+const config = cli(defaultConfig);
 
 if (config.debug) {
   process.env.DEBUG = "oasis,oasis:*";
 }
-
-console.log(JSON.parse(JSON.stringify(config)));
 
 process.on("uncaughtException", function(err) {
   // This isn't `err.code` because TypeScript doesn't like that.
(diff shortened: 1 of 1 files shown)


black-puppydog:
@christianbundy no, I'm fine with `default.json`. I think the `.config/` part makes it clear what's happening.
But the `console.log` was actually there on purpose. I recall that it is often one of the first lines of inquiry for any support work in patchwork. Of course, the json.endoce/decode is pointless...

black-puppydog:
btw thanks for this. reading your code is very welcome feedback :)

christianbundy:
> But the `console.log` was actually there on purpose. I recall that it is often one of the first lines of inquiry for any support work in patchwork. Of course, the json.endoce/decode is pointless...

For me, the `console.log` was just printing like this:

```console
$ oasis
Trying to read config file /home/christianbundy/.config/oasis/config.json
{
  _: [],
  open: true,
  offline: false,
  host: 'localhost',
  port: 3000,
  public: false,
  debug: false,
  '$0': 'oasis'
}
```

I don't think that's what we want, right? I think we're already logging most of the relevant stuff, like the host + port + offline, and debug can be inferred, but maybe we could add messages for the others? 

If this is something that's important to you I'd be open to discussing printing some config output for debugging in an issue, but I *am* hesitant to print a wall of text like how Patchwork does because I think it conditions people to just ignore what's happening in their terminal.

christianbundy:
> btw thanks for this. reading your code is very welcome feedback :)

Thank **you**! I'm really happy that we can change the defaults now, I hadn't come across the problem before but I can see how great of a solution this is.

cinnamon-bun:
Thanks @black-puppydog and @christianbundy !

I added an issue to document the config file and make it discoverable: #254

christianbundy:
Thanks or merging and adding the issue! :+1:

black-puppydog:
> I don't think that's what we want, right

Actually that was exactly what I wanted (minus the `_` and `$0`) because it allows to see what the current configuration (for oasis, not for ssb-server) is, in one place. 
I guess @christianbundy you were thinking of the ssb-server config as we also show it in patchwork? That might indeed also be helpful. But I actually like the fact that now there's a separate config for the ssb *server* and the ssb *viewing application*. Before, everything was lumped together in `~/.ssb` and it was a mess...

christianbundy:
I think that would be really useful in debug mode, maybe we could add it there? I was thinking of the SSB-Config that gets printed before it's used to create an SSB service, yep.

black-puppydog:
From how I read the [patchwork code](https://github.com/ssbc/patchwork/blob/eb9223ce8855c3038e71cd491ab9233aef852212/index.js#L236), what we print there is actually the config as used *after* all injection and modification. So an aggregate of built-in defaults and `~/.ssb/config` (I don't see any CLI parameters there) as used further down in the code. 
So pretty much what we're doing in Oasis now: generate the full, effective config, and print that.

PR #257, opened 2020-02-24
Update README - fix config path on mac os

cinnamon-bun:
Correct the config path on Mac Os.


cinnamon-bun:
Usually we make PRs from our fork repos, but this is such a small one that I did it from the Github web editor, which saves a lot work (otherwise I wouldn't have had time to do this right now).

Are we comfortable with small exceptions to our process like this, or would we like to be more strict?

christianbundy:
Thanks for this patch!

> Are we comfortable with small exceptions to our process like this, or would we like to be more strict?

Very comfortable with small exceptions!

PR #262, opened 2020-02-26
Detect existing Oasis server running on interface

christianbundy:
Problem: Sometimes you try to run Oasis twice and the second time you
run the program it blows up in your face about the port being taken.
Most other software just focuses or reopens the window instead of
throwing errors at you.

Solution: Implement `/.well-known/oasis` as a sanity check to see
whether the thing listening on the HTTP port. If it replies "oasis" then
we can open the tab (unless `config.open === false`) and avoid throwing
the error.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,11 +7,20 @@
 const fs = require("fs");
 const envPaths = require("env-paths");
 const path = require("path");
+const nodeHttp = require("http");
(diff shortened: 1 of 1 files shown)


cinnamon-bun:
This is much more user-friendly!  Great idea

PR #263, opened 2020-02-26
Hide search results from blocked feeds

christianbundy:
Problem: The search results weren't hiding results from blocked feeds.

Solution: Add `socialFilter()` to ensure that all blocked feeds are
removed from search results as the first step in the stream processing.


cinnamon-bun:
Yay!

I wonder how we can test this more rigorously since there are so many places blocked feeds can show up.  Maybe with some test accounts, a test network, ?

PR #269, opened 2020-02-27
Add hashtag support to search

christianbundy:
Problem: Searching for a hashtag should bring you to the hashtag page,
and the hashtag page should have some useful information about which
page you're on instead of just showing you the messages.

Solution: Add code so that if you search for a hashtag you're brought to
that page, and display some useful help text at the top of the page
explaining the hashtag page. While ensuring that the hashtag page showed
posts from around the network (instead of only people you're following)
I fixed the popular page so that it has the same behavior again (fixing
a regression) and filtered out gatherings (fixing *another* regression)
on the popular page. These probably should've happened in another commit
but I got carried away. :/

Happy to split this commit into two if it hurts to bundle the popular
fixes with the hashtag fix, but I'm low on energy and want to at least
open a PR in case it doesn't bother anyone.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -121,7 +121,8 @@ const {
   extendedView,
   latestView,
   likesView,
-  listView,
+  threadView,
+  hashtagView,
   markdownView,
   mentionsView,
   popularView,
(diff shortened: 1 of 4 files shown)


cinnamon-bun:
Thanks!

I wouldn't mind seeing gatherings in most of the feeds since they can be part of threads.  ...once we can render them nicely, anyway.

PR #276, opened 2020-03-01
Simplify installation

christianbundy:
Problem: The install instructions in the readme contains quotes because
it has a `*`, but `#semver:` does what we need without the quotes. The
`docs/install.md` file also has some unnecessary complexity, like
cloning via SSH (only useful for maintainers), which I think we can
safely remove.

Solution: Change the install instruction and reorganize
`docs/install.md` to be more relevant to people who are installing from
source.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -3,6 +3,7 @@
   "language": "en",
   "words": [
     "AGPL",
+    "EACCESS",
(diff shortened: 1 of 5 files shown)


christianbundy:
Thanks!!

PR #277, opened 2020-03-01
Add profile image upload

christianbundy:
Problem: We can set our name and profile description but profile images
are expected by most people and supported by most clients and we don't
have them. A profile without an image can sometimes lack the intimacy
you'd get if you let people upload profile images that they can use as a
visual avatar.

Solution: Add profile image upload to the Edit Profile page and add a
bunch of plumbing for `ssb.blobs.add()` to add the blob and publish a
message setting it as a profile image.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -21,6 +21,8 @@
     "libsodium",
     "libtool",
     "manyverse",
+    "mebibyte",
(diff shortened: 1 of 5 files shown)


cinnamon-bun:
✨ 🖼 ✨

PR #279, opened 2020-03-02
README.md: clarify installation instructions

cinnamon-bun:
The example command `oasis` all by itself was so simple that it wasn't clearly a command-line command, so I added a sentence about it.

Also added a sentence about installing Node, geared at non-technical people.


christianbundy:
This is wonderful, thank you for improving the readme!

PR #281, opened 2020-03-02
README: add FAQ about running alongside other SSB apps

cinnamon-bun:
Patchwork freezes on launch if Oasis is already running (because of an SSB server conflict, I assume).  I added an FAQ to guide people around that pothole.


christianbundy:
Thanks for this patch! Thanks for being so patient waiting for a merge, here we go... :rocket:

PR #282, opened 2020-03-03
Updated Github issue and pull request templates

georgeowell:
This could be just personal preference but I prefer using markdown headers for the template questions. I also thought it would be useful to ask what version people are using for issues.

Updated the templates.

P.s. I am also using this as a chance to test GPG signed commits :smile:


christianbundy:
LGTM, thanks for this patch!

PR #283, opened 2020-03-03
Fixed wrapping in README to improve readability of /settings/readme

georgeowell:
When viewing /settings/readme as linked in the setting page, there was some issues with the way README was formatted as you can see in this screenshot:

Screenshot_2020-03-03 Oasis

Tweak markdown so that it looks better in the browser:
Screenshot_2020-03-03 Oasis(1)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,18 @@
 # Oasis
 
-Oasis is a **free, open-source, peer-to-peer social application** that helps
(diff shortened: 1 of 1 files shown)


christianbundy:
Looks like it resolves the newline problem, thanks for the patch. :+1:

Issue #220, opened 2020-02-16
Hide forks/replies until the user expands them


Powersource:
More context in this thread `%YiBecWPnG9iaftO6xRbzsUYxjYOBuZCBI5IEVnH/Uc4=.sha256` , copy pasting some of my thoughts from there to here

**What's the problem you want solved?**

> One issue with [reddit-style thread formatting] that I noticed when trying [oasis] out is that someone made a fork, I assume to not disturb the “main” thread. But since that fork thread was embedded in the main thread it instead took up a lot of space, and if you wanted to read the next message in the main thread you had to scroll all the way past the fork thread.

**Is there a solution you'd like to recommend?**

> One solution could maybe be to by default only show a small preview of the fork, a bit like patchwork, but when you click on it it expands inline and ends up looking sort of like your screenshots above (with maybe a button for un-expanding it again as well). Does that sound like something you’d merge? (not saying I’m gonna implement it :p )

christianbundy:
Thanks for bringing this up! I think this is the right move and I think `<details>` and `<summary>` would be a great way to implement this.

Powersource:
Nice, I was wondering if we could do this without js.

And I just tested, it seems possible to nest `<details>` 🤯

Powersource:
HOLY SHIT I think I made it work?? Pushing in a min

Powersource:
tada https://github.com/fraction/oasis/pull/292

PR #292, opened 2020-03-06
Fix "Hide forks/replies until the user expands them"

Powersource:

Fixes https://github.com/fraction/oasis/issues/220

This one :D Or well this but a bit cleaned up, in this first commit I've included my failed attempts for posterity. You could also argue that this is pretty hacky and should be rewritten lol, and I'd partially agree but tbh I don't know if the other solutions  would be much cleaner.

Good thread for testing (many forks) http://localhost:3000/thread/%25YiBecWPnG9iaftO6xRbzsUYxjYOBuZCBI5IEVnH%2FUc4%3D.sha256

## Todo

* [x] Remove old code and comments
* [x] Make sure all views are rendered with this, I think atm it might just be the specific thread view but idk
* [x] Add `<summary>`ies, using e.g. the first part of the first reply in the fork.
* [x] Make it a bit prettier/more understandable. E.g. right now the `Details` text has 0 left margin, and I'd also like a border on the left of the `<details>`. And since the posts are rendered inside the Details we could move the fork-posts' left margin to the Details.
* [x] If a message in a fork is linked to, the fork should be pre-expanded


christianbundy:
:tada: Thanks for this patch! I've got a friend visiting from out of town today so I might be slow to check it out, but I'm really excited to have this functionality.

Powersource:
Yeah no pressure, I've got a bunch of changes needed before I'm happy with this :)

Powersource:
Ran into a fun thing where the typescript checker tricked me into removing `.flat()` in one spot because I thought it had figured out it wasn't needed, but that's apparently because we're using an ["old" version of ts](https://stackoverflow.com/questions/53556409/typescript-flatmap-flat-flatten-doesnt-exist-on-type-any). Anyway fixed that by just using `lodash.flatten` instead.

christianbundy:
FWIW I don't think Node 10 has `Array.prototype.flat()` either way.

Powersource:
Yeah I think the issue was pretty much that, I forgot that we're supporting LTS.

Powersource:
a screenshot, now that there's some styling so it's a bit more understandable what's happening

image

Powersource:
image

christianbundy:
Just took this for a spin and it works really great. The thread view and Summaries view are the only ones with nesting, so we're good to go from that angle. The only big todo is making sure that we auto-expand all parent `<details>` when linking to a message.

christianbundy:
Please let me know when you'd like me to take a look at this again! Since it's an open PR I get a notification each time you push, but I'm not sure when it's ready for review.

Powersource:
Yep should be ready for review now!

Powersource:
Wow that was quick! Nice to have this in :)

Was wondering if you had opinions on the preprocessing step/`lodash.set(...meta...)` being outside of models.js?

christianbundy:
Awesome! Thanks a bunch for this patch. I'm going to add you as a maintainer. Please feel free to check out the [development process](https://github.com/fraction/oasis/blob/master/docs/contributing.md#24-development-process), and either create an issue or reach out if you're not sure about something.

Welcome aboard. :train: :train:

Powersource:
Thanks! And tbh I'm not gonna be able to remember that whole process document but I'll try :sweat_smile: Just tell me if something's off.

Also bumping my q since I'm guessing you missed it when we commented at the same time :)  
> Was wondering if you had opinions on the preprocessing step/lodash.set(...meta...) being outside of models.js?

christianbundy:
TL;DR: Please merge any branch that passes existing tests and solves a problem, except when it causes another problem.

> Was wondering if you had opinions on the preprocessing step/lodash.set(...meta...) being outside of models.js?

As a maintainer I don't have opinions on implementation details, but as a contributor I think I'd probably like to see that sort of thing in the models. I've been thinking that I'd like to write a patch to refactor the models so that they return nested arrays of threads rather than `depth`, that way we don't need to handle this in the views, but I haven't had the time or energy to write that patch yet. If you have thoughts, I'd be super interested to hear them!

christianbundy:
And yep, I totally missed your question, thanks for pointing it out. :100:

PR #293, opened 2020-03-08
Add German translations

hng:

Missing German translations.

I've added German translations for all English strings in i18n.js

## Still open

* Some strings are not translatable right now: e.g. the label of the content warning text field
* I don't know what the right way is to translate variables, e.g. the "public/private" part in "write a public/private comment" when writing a comment.


christianbundy:
This is awesome, thanks for the patch!

PR #295, opened 2020-03-09
Small i18n improvements

### LINKED ISSUE #294: Not translatable

From @hng in #293:

> - Some strings are not translatable right now: e.g. the label of the content warning text field
> - I don't know what the right way is to translate variables, e.g. the "public/private" part in "write a public/private comment" when writing a comment.

hng:

Some strings were still hardcoded and not translatable. Fixes #294 

Added a i18n string for the contentWarningPlaceholder and for "private/public" adjectives when writing a comment.

## Questions

I am not super confident about my naming of the i18n strings, so I am open to other naming suggestions.

### DIFF
--- a/src/views/i18n.js
+++ b/src/views/i18n.js
@@ -75,6 +75,8 @@ module.exports = {
       strong("reply"),
       " instead."
     ],
+    commentPublic: "public",
+    commentPrivate: "private",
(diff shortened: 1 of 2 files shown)


christianbundy:
This looks great, thank you for fixing this! :heart:

hng:
No problem. Thank you for all your work! :)

PR #299, opened 2020-03-09
Present error message while waiting for indexing

christianbundy:
Problem: When our views are still indexing the database they apparently
don't respond over MuxRPC, which means that we're just waiting forever
until they finish. This means that people who are indexing the database
get an HTTP response that might take an hour to finish.

Solution: If we have more than 1 mebibyte of backlog, present an error
message explaining the situation and asking for a bit of patience.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -22,6 +22,7 @@
     "libsodium",
     "libtool",
     "manyverse",
+    "mebibyte",
     "minlength",
     "mkdir",
     "monokai",
--- a/src/http.js
+++ b/src/http.js
@@ -16,8 +16,8 @@ module.exports = ({ host, port, middleware }) => {
   app.on("error", err => {
     // Output full error objects
     err.message = err.stack;
-    err.expose = true;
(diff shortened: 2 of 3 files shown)


cinnamon-bun:
Awesome fix!

For future reference, here's what it looks like (in browser and CLI):
```
Error: Sorry, Oasis has only processed 95.4% of the messages and needs to catch up.
       Thanks for your patience, please wait for a moment and refresh this page to try again.
    at middleware (/Users/x/projects/ssb/oasis-github/src/index.js:776:13)
    at processTicksAndRejections (internal/process/task_queues.js:94:5)
    at async /Users/x/projects/ssb/oasis-github/src/index.js:756:5
    at async /Users/x/projects/ssb/oasis-github/src/index.js:751:5
    at async /Users/x/projects/ssb/oasis-github/src/http.js:28:5
    at async /Users/x/projects/ssb/oasis-github/node_modules/koa-mount/index.js:52:26
```

PR #305, opened 2020-03-11
Lands on mentions

seekr:
Since we don't have notifications in Oasis yet, it may be useful to have Oasis land on /mentions instead of /public/popular/day. This way user can always start the journey by checking for new mentions.

I hard coded the new root redirect to /mentions. A next step could be to create a new setting allow user to pick what should be the "home" tab

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -3,6 +3,7 @@
   "language": "en",
   "words": [
     "AGPL",
+    "EACCESS",
     "Argyris",
     "Hintjens",
     "Kata",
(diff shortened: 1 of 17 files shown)


christianbundy:
Oh no, it looks like maybe you forked my personal repo which has lots of experimental code in it. I'll try to cherry-pick your commit. You can fix it locally with:

```sh
git remote set-url origin git@github.com:fraction/oasis.git
git fetch origin
git checkout master
git reset --hard origin/master
```

christianbundy:
Merged in #306.

PR #309, opened 2020-03-11
Fix link to popular view in navigation

hng:

The link to the popular view in the navigation was broken after changing the root to /mentions.

Added direkt link to popular view instead of root.


christianbundy:
Oh, thanks for catching this! This is great.

PR #311, opened 2020-03-11
Rearrange settings

georgeowell:
Settings list is in a bit of an arbitrary order.

Rearrange settings list in order of potentially most likely to use.


christianbundy:
Beautiful, thank you!

Issue #288, opened 2020-03-04
AUTHORS and MAINTAINERS only have one person


christianbundy:
## What's the problem you want solved?

This project has lots of contributors and a few maintainers but the files don't have anyone else's info. It would be cool to give credit to everyone who's poured work into the project! I added myself to these files just to start them off, but feel kinda weird about being the only person in the files.

## Is there a solution you'd like to recommend?

I kinda want to just add people but I'm not sure whether that's rude. I think the C4 doc says something about authors being responsible for adding themselves to the file, but it feels weird to add that responsibility.

Alternatively we could delete these files? I think keeping track of maintainers is important so that people can email us security problems and stuff, but the only reason I want an authors file is to celebrate all the cool folks contributing patches. It's possible that Git metadata is enough for that, but if we use that we should use [mailmap](https://www.git-scm.com/docs/git-check-mailmap) so that our friends can choose their canonical names. I'd feel real bad about deadnaming someone forever through Git metadata. 😱

## What version or commit of Oasis are you using?

latest

PR #312, opened 2020-03-11
Add myself to MAINTAINERS

cinnamon-bun:
#288 AUTHORS and MAINTAINERS have only one person :)

I feel like a MAINTAINER.  I don't feel like an AUTHOR so I'm not adding myself there - I haven't written much actual code - but if it was renamed to CONTRIBUTORS I would add myself there also.


christianbundy:
Beautiful!

PR #316, opened 2020-03-12
Small text tweaks

georgeowell:
Posts from yourself show up in certain feeds now. This wasn't mentioned in the feed descriptions.

Added that reference to the descriptions.

### DIFF
--- a/src/views/i18n.js
+++ b/src/views/i18n.js
@@ -22,12 +22,12 @@ module.exports = {
     topics: "Topics",
     topicsDescription: [
       strong("Topics"),
-      " from people  you follow, sorted by recency. Select the timestamp of any post to see the rest of the thread."
+      " from yourself and people you follow, sorted by recency. Select the timestamp of any post to see the rest of the thread."
     ],
     summaries: "Summaries",
     summariesDescription: [
       strong("Topics and some comments"),
-      " from people  you follow, sorted by recency. Select the timestamp of any post to see the rest of the thread."
+      " from yourself and people you follow, sorted by recency. Select the timestamp of any post to see the rest of the thread."
     ],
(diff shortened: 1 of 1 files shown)


georgeowell:
Hmm, the CI failed :cry:

christianbundy:
@georgeowell `npm run fix` to the rescue!

georgeowell:
@christianbundy what does that do?

christianbundy:
That runs the "fix" script defined in `package.json`, which runs `common-good fix`, which makes [Common-Good](https://github.com/christianbundy/common-good) automatically fix any problems it can with [Stylelint, Prettier, and ESLint](https://github.com/christianbundy/common-good/blob/6077dd89ff61dfccc4d731d7fa79a88a05a7a204/index.js#L39-L41).

cinnamon-bun:
Thanks @georgeowell 🌟 

Side note, I've found the CI checks to be a barrier to my contributions.  Sometimes I want to make changes only using the GitHub UI, not the command line, and those always seem to fail.

Maybe we could exempt certain files from this check?  Files that non-coders might like to change, particularly Markdown files and `i18n`.

I think that would be in the spirit of the C4 process, to accept PRs with formatting imperfections.  We can always run `common-good fix` after merging them.

(I'm not opposed to autoformatting, just don't want to block merges because of that, for certain files)

christianbundy:
Yep, I'm happy to run `npm run fix` in the merge commit, I've just been putting it off. Thanks for all your patience @georgeowell!

Here's what I'm doing locally:

```sh
git pull upstream pull/316/head
npm run fix
git add -A
git commit --amend --no-edit
git push upstream master
```

PR #317, opened 2020-03-12
Added reference to "plain text" instead of just text

georgeowell:
Text is ambiguous. Perhaps the full term "plain text" is better when compared to "Markdown"

Added reference to "plain text" instead of just text


christianbundy:
Super reasonable, thanks for this patch!

PR #324, opened 2020-03-17
An html error message.

### LINKED ISSUE #323: meta refresh on processing errors
Right now, I get an error about the system not being done processing. "Sorry, Oasis has only processed xx.x% of the messages and needs to catch up". I'd like to know when it's done.

It would be nice if the error page that's rendered in my browser had a [meta refresh tag](https://en.wikipedia.org/wiki/Meta_refresh) which would reload it until it worked.

2.14.0

justinabrahms:
This isn't an actually good patch, but it's something that actually returns html errors.

You should merge this (or copy it) and use this to fix #323.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -774,9 +774,10 @@ const middleware = [
     const percent = Math.floor((totalCurrent / totalTarget) * 1000) / 10;
     const mebibyte = 1024 * 1024;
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Thanks! I've just opened #325 to push this along.

PR #335, opened 2020-03-20
Fix crash when linking directly to private message

christianbundy:
Problem: Trying to view a private message crashes the server and leaks a
bunch of memory. That's bad. This problem is caused by faulty handling
in the function that finds thread ancestors. There's some code that says
"if the next ancestor is a private message, just return the ancestors
that we know about", which returns nothing when we're looking up a
private post (because we can't identify *any* ancestors).

Solution: Ensure that we're only resolving the promise once in the
function by chaining `else if`s and ensure that we only return the
ancestor list if there are actually ancestor in it. If the ancestor list
is empty, we can just return the single message that we know about and
pass it off.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -1052,17 +1052,22 @@ module.exports = ({ cooler, isPublic }) => {
                 debug("getting root ancestor of %s", msg.key);
 
                 if (typeof msg.value.content === "string") {
-                  debug("private message");
                   // Private message we can't decrypt, stop looking for parents.
-                  resolve(parents);
-                }
-
-                if (msg.value.content.type !== "post") {
+                  debug("private message");
+                  if (parents.length > 0) {
+                    // If we already have some parents, return those.
+                    resolve(parents);
+                  } else {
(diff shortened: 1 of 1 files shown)


christianbundy:
@soapdog is this something you might have time to take a look at?

src/models.js
```diff
@@ -1052,17 +1052,22 @@ module.exports = ({ cooler, isPublic }) => {
...
-                  debug("private message");
                   // Private message we can't decrypt, stop looking for parents.
-                  resolve(parents);
-                }
-
-                if (msg.value.content.type !== "post") {
+                  debug("private message");
+                  if (parents.length > 0) {
+                    // If we already have some parents, return those.
+                    resolve(parents);
```
sbillig: Is it ever appropriate to resolve to an array? Everything seems to assume the resolved value is a msg.

Typescript plz.

christianbundy:
This has also been open for a few days without review, self-merging.

PR #337, opened 2020-03-22
Actually shift cards over, rather than just a minor indent.

### LINKED ISSUE #336: Improve indentation for threads
I find it hard to visually track threads as separate from posts. This is especially true if there's any sort of nesting (which I think I've seen??)

Have responses threads be offset a little so the right side is visually distinct from the others.

Example:

more-offset-oasis-2

2.14.0

justinabrahms:
Previously, there was a minor indent on threads, which made it visually difficult to determine which things were replies and which were new posts. 

This addresses that concern by shifting the individual cards over, which makes them visually distinct. 

Additionally, it doesn't move the sidebar over at all, so there's no weirdness as you navigate tabs.

Fixes #336

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -517,6 +517,6 @@ hr {
  * be wise to nest these recursively on the thread view, which would make it so
  * that we don't need any inline CSS anymore.
  */
-.indent {
-  margin-left: 1rem;
+.indent > section {
+  margin-left: 2rem;
 }


christianbundy:
Thanks for this patch! Travis is being weird and saying "An error occurred while generating the build script.", but everything looks good to me. I'm going to give Travis like 24 hours before mucking around in the settings to merge this.

Btw, we don't need to resolve this problem in this PR, and maybe this isn't even a problem? But this option (overflowing the `<main>` container) has a trade-off on mobile where it adds horizontal scrolling:

Screenshot from 2020-03-22 07-44-03

Maybe that's okay? :man_shrugging: Happy to merge this as-is once Travis starts working again, but I just wanted to make a note of the change.

justinabrahms:
Good call out. I didn't test this on mobile. I suspect we can do something slightly different on mobile like color the left offset a bit and use a margin on the .indent div to accomplish similar needs.

PR #340, opened 2020-03-23
Update npm dependencies

christianbundy:
Problem: Our dependencies are falling behind and it'd be nice to have
them up-to-date. The only big update it that Common-Good upgraded to the
latest version of Prettier.

Solution: Update them!


thorn0:
It's `prettier --write`, not `--fix`.

thorn0:
`&& git diff` → `|| git diff`

christianbundy:
Oh, thanks. I assumed that `prettier --write .` would also exit 0, but maybe it has errors too. I'll change to your suggestion once this commit fails like the others. Sorry, my brain's a bit fried.

thorn0:
Um, you're actually right. `--write` should return 0.

thorn0:
Try the `file` command, from https://stackoverflow.com/a/3570574/76173 :
```console
$ file testfile1.txt
testfile.txt: ASCII text

$ file testfile2.txt
testfile2.txt: ASCII text, with CRLF line terminators
```

christianbundy:
Resolved the above problems in https://github.com/prettier/prettier/issues/7853. :tada: Thanks for your help!

package.json
```diff
@@ -22,7 +22,8 @@
   },
   "dependencies": {
     "@fraction/base16-css": "^1.1.0",
-    "@fraction/flotilla": "^4.0.1",
+    "@fraction/flotilla": "^4.0.0",
```
jedahan: flotilla downgrade?

christianbundy: Oops, unintentional. I'll fix that now.

christianbundy:
@jedahan fixed! please merge when you're happy with this!

jedahan:
once the conflict is fixed i'll be happy to
though i cannot test it, the updates look fine

christianbundy:
Done! Ready to go now.

christianbundy:
Thanks for the merge!

PR #341, opened 2020-03-24
Don't append channel to non-root posts

christianbundy:
Problem: When publishing a root post, other clients give you the option
of adding a `channel` property, which is basically just a hashtag. We
show this at the end of the message, but since replies often copy the
`channel` property then it gets appended to every reply.

Solution: Only append the channel as a hashtag for root posts.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -466,10 +466,17 @@ module.exports = ({ cooler, isPublic }) => {
           }
         }
 
+        const isPost =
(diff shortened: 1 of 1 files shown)


sbillig:
Looks good; might be reasonable to show the channel for non-root posts when displaying them outside of a thread context (eg. on the Popular view), but that's a change for another day.

PR #342, opened 2020-03-24
Only show connected peers in settings

christianbundy:
Problem: The peer list shows peers that are connected and in the process
of establishing a new connection, which has lots of turbulence. It's
confusing to see dozens of "connections" that only exist for a few
milliseconds when they're actually just connection attempts.

Solution: Only show peers with the state "connected".


christianbundy:
Thanks!

PR #343, opened 2020-03-24
Load theme using existing config mechanisms

### LINKED ISSUE #333: Oasis doesn't remember my preferred color scheme.
I've noticed that Oasis switches back to the default color scheme every time I launch it (I usually like setting it to `monokai`). It would be cool if my choice of color scheme persists between sessions.

Maybe whenever someone sets their preferences through the settings menu in oasis, these settings get saved in the `default.json` file.

2.14.9

nickwynja:

Closes #333 

By using the existing config mechanism, this allow someone to pass in the theme as a command line arg or set the theme in `defaults.json`, while still allowing someone to set the theme on a per-browser basis as a cookie.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -43,4 +43,9 @@ module.exports = (presets, defaultConfigFile) =>
       default: _.get(presets, "debug", false),
       type: "boolean"
     })
+    .options("theme", {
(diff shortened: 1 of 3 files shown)


christianbundy:
This is wonderful, thank you for the patch!

PR #345, opened 2020-03-24
Remove version question from issue template

christianbundy:
Problem: Almost everyone is on the latest version and we probably don't
need to ask for what version they're using.

Solution: Remove the question.


christianbundy:
Woo!

PR #348, opened 2020-03-24
Text-based browser spacing issues and sticky headers enhancement

### LINKED ISSUE #329: Spacing issue in text based browsers
There are some issues with spacing between items when using Oasis in a text based browser. I tried Lynx.

Screenshot from 2020-03-19 02-31-53

As you can see, all the post action options at the top of each post ("CommentJSON...) is all smushed together. 

Ensure spacing between post action options are maintained.

Latest

nickwynja:

Addresses spacing issues in text-based browsers by introducing block elements and non-breaking spaces. Closes #329 

Also introduces sticky headers. While reading long conversations, it's easy to get a bit lost as to whom is saying what. With some basic CSS, we can keep the header of the post sticky so you can see more context above the message.

Apologies about combining these two into one PR. I actually started with the sticky headers before realizing that some of the block elements for the text-based browsers would help with the sticky feature.

***

(Submitting as Draft since I think I need to test this more in both browser types, as well as get feedback on the idea of the sticky headers in general.)


christianbundy:
Looks great to me, the text fixes are super appreciated and I think the sticky header is a step in the right direction. I've tested in Firefox and Lynx and it looks lovely, let me know when you'd like to see this merged. :tada:

nickwynja:
@christianbundy I think this is all set now. Fixed up a couple minor things on the Profile page and with the section headings on a few pages.

christianbundy:
This looks wonderful, thanks for sharing this patch! If you bump into pain points I hope you'll let me know, lots of this code came from the original prototype and I know it's got lots of room for improvement. Thanks for doing the work.

nickwynja:
(My reply from email isn't appearing so re-posting below.)

***
You're welcome! Glad it's useful.

I'm pretty new to writing features in node apps so it could be a bit of a learning curve but having some experience with python flask, Angular apps, and more recently Vue, I feel like HTML templates are much more approachable and simple in the long run. It took some digging to find out how the HTML is rendered before I discovered `hyperaxe`. Depending how Oasis grows, I can see templates and partials being much more manageable.

I'm not sure whether it's a preference or what's most simple with `hyperaxe` but I was running into some complications with styles being tied to elements and their structure, like `section > header`. In past projects, I've seen this become messy quickly and preferred using elements for semantic meaning combined with classes for styling. For example, `section > footer > form > button.liked` could simply be something like `.post button.liked` to drop the deep nesting, share styles across structures, and allow for the structure of the component HTML to evolve as needed. 

That said, it was pretty easy for me to get into it and in general just love how lightweight Oasis is. Makes for a good browsing experience and a good hacking experience!

PR #349, opened 2020-03-24
Fix theme using capitalization

christianbundy:
Problem: Capitalization in "sulphurPool" was causing a file not found.

Solution: Repalce "sulphurPool" with "sulphurpool".


sbillig:
Looks right to me. (I don't think I can merge)

christianbundy:
Thanks for the merge!

PR #350, opened 2020-03-24
Add .mailmap for canonical identities

christianbundy:
Problem: Some of us have a few different emails, which show up as
different people when we commit. It would be great to merge these Git
identities so that they don't show up as different people.

Solution: Use `.mailmap`, which lets people choose their favorite
canonical email address for this project.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,4 @@
+Cinnamon <cinnamon_oasis@fastmail.com> <cinnamon_bun_github@fastmail.com>
+Cinnamon <cinnamon_oasis@fastmail.com> <32660718+cinnamon-bun@users.noreply.github.com>
+Jonathan Dahan <github@jonathan.is> <hi@jonathan.is>
+kawaiipunk <kawaiipunk@posteo.net> <georgeowell@users.noreply.github.com>


christianbundy:
cc: @cinnamon-bun @jedahan @georgeowell please let me know if I've chosen your primary email address incorrectly and I'll write a patch to fix it.

christianbundy:
@jedahan please merge when you want!

(btw the required approvals were disabled a bit ago, so you should be able to just merge instead of going through the code review approval process)

jedahan:
For this I would like to wait for @georgeowell and @cinnamon-bun since it has to do with identities

georgeowell:
I don't know what I'm doing on this strange Microsoft platform. kawaiipunk@posteo.net is fine for both.

christianbundy:
Cool, and Cinnamon added their own email [here](https://github.com/fraction/oasis/commit/363d4f869b34ec8c88e59fcc0f435a5394fdc370), but we can also wait. The other open PRs are much higher priority anyway, so no big deal if we need to wait on this one.

sbillig:
Seems like everyone's cool with the listed emails; merging.

PR #352, opened 2020-03-25
Fall back to TCP socket if Unix connection fails

christianbundy:
Problem: We're only attempting connection via Unix sockets, but some SSB
services may choose to only listen on TCP sockets.

Solution: Start by attempting connection to Unix socket and fallback to
the TCP socket remote if that doesn't work, before just trying to start
our own SSB service.


christianbundy:
cc: @khimaros could you confirm that this fix works?

khimaros:
@christianbundy -- great catch! This had nothing to do with caps after all.

Initially this didn't help because my `net` connection was not configured to listen on all interfaces (maybe oasis was looking for `127.0.0.1`?).

After setting `"host": "0.0.0.0"`, so that it listened on all interfaces, it worked. An alternative solution was adding `"unix": [ {"scope": "device", "transform": "noauth" }]`.

christianbundy:
Thanks for confirming!

This *should* be using the `remote` property from your config, but if you don't have one then SSB-Client has [some defaults](https://github.com/ssbc/ssb-client/blob/e2ff27019657916cd13c8fb7a840f2675eaf4a1d/index.js#L31-L34), which would explain the problem.

khimaros:
Though it isn't a panacea, this seems worth merging.

sbillig:
Finally took the time to test this. Seems to work fine.

Issue #249, opened 2020-02-20
Merge Summaries concept into Topics and Popular


georgeowell:
## What's the problem you want solved?
Summaries is pretty decent now. I just wondered if anyone wanted to merge that visible comments concept into Topics and Popular. 

This would be a much better experience in navigating public threads as it allow you to see that a post has become a discussion, a preview of the nature of discussion and also how many comments exist. 

In addition, when combined with #179, we would have a clearer label to click through to the thread rather than clicking a unrelated piece of information (how long since the post was made).

I also wish the visible comments were limited to two rather than three as it takes up quite a bit of screen space on my laptop.

## Is there a solution you'd like to recommend?
- Removing summaries view
- Adding comment summaries to Topics view
- Adding comment summaries to Popular view

christianbundy:
I'd merge this but I think it would be more prudent to make filters or something like in #170 so that we can switch between the raw view and the summary view everywhere.

georgeowell:
> I'd merge this but I think it would be more prudent to make filters or something like in #170 so that we can switch between the raw view and the summary view everywhere.

Fair enough! Perhaps https://github.com/fraction/oasis/issues/179 would be a good step for now then?

georgeowell:
I think perhaps Topics and Summaries should just be merged into Threads. Threads seem like a good compromise for what those views were trying achieve.

PR #353, opened 2020-03-25
Introduces Threads

nickwynja:

I found it difficult to discover active conversations in my network. I'd see replies to posts in "Latest" or find new people in "Extended" but no way to discussions. Threads introduces a way to see threaded conversations of both people that you follow and that are in your extended network.

Even though there are already several ways to display messages by time or conversations, there wasn't yet _this way_. "Threads" shows posts from both people you follow and people in your extended network that have comments or replies and shows a preview of some of the thread. This is a good way to explore active conversations that are happening in your network.

Relates to #249, #220, #170. I tried to think through how I could merge this view into existing views, similar to how Popular has the time selection options in the view. I considered adding this as a view within Summaries but wasn't sure of a way to do that well. This implementation duplicates most of the code from Summaries view, which isn't ideal either. I figured I'd get this view out there and then could work out where it ends up.


christianbundy:
Looks good, I think it just needs a `npm run fix` and it's ready to merge. In the future I think it'd be cool to have one "Public posts" view that has all of the view / sort options so that we don't have a combinatorial explosion, but for the time being I think this is totally fine.

christianbundy:
Thanks again!

PR #357, opened 2020-03-25
Fix UI inconsistencies

christianbundy:
Problem: After some fantastic new pull requests, there are some
inconsistencies between how each author implemented different features.

Solution: Fix regressions and iron out inconsistencies. This fixes the
heart highlight bug, uses the same indentation for Summaries + Threads +
the thread view, tightens up the CSP now that we don't need
unsafe-inline styles, and uses a neutral tone for indents with the same
width as the blockquote border (so far violet has been reserved for
private messages).

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -196,13 +196,14 @@ textarea:focus {
   border-color: var(--blue);
 }
 
+/* Prevent button styles being applied to heart button */
 button:focus,
 button:hover {
   background-color: var(--fg-light);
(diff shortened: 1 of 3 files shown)


nickwynja:
@christianbundy Thanks for cleaning up some of my mess! 😬😂

christianbundy:
Thanks for merging this! Your patch was great, a little turbulence never hurt anyone. :1st_place_medal:

PR #360, opened 2020-03-26
fix follow button link for non-english languages

msiqu:
currently if non-english language is set in configuration, the FOLLOW button doesn't work (it takes the i18n translation as part of the URL).

made the url fixed `*/follow/*` rather than dynamic

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -631,8 +631,8 @@ exports.authorView = ({
     relationship.following === true &&
     relationship.blocking === false;
 
-  const contactFormType = areFollowing ? 'unfollow' : 'follow';
-  const contactFormTypeLabel = areFollowing ? i18n.unfollow : i18n.follow;
+  const contactFormType = areFollowing ? "unfollow" : "follow";
+  const contactFormTypeLabel = i18n[contactFormType];
 
   const contactForm =
     relationship === null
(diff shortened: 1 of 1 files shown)


christianbundy:
This is great! But does this also work when you're already following someone? In that case, it should instead **un**follow.

msiqu:
goog question

> This is great! But does this also work when you're already following someone? In that case, it should instead **un**follow.

Good point ... it does not fix the unfollow... (I guess I don't need to unfollow anyone! ;D) Let me keep thinking how to fix this...#notAcoder

msiqu:
@christianbundy so, to my untrained eye [my fix](https://github.com/fraction/oasis/pull/360/commits/2c86eb5cf18afd191018381d63f9f7169817b017) solves the problem but it feels dirty. What do you think?

christianbundy:
Looks like a fix to me! I'd be happy to merge this.

If you really want to, you could do something like this:

```javascript
const contactFormType = areFollowing ? 'unfollow' : 'follow';
const contactFormTypeLabel = i18n[contactFormType];
```

But this is a great patch and I'm happy to merge it as-is.

christianbundy:
Oops, looks like `npm test` failed -- could you try running `npm run fix` and commit the change? Sorry, the linter is a bit picky.

msiqu:
@christianbundy cleaned... sorry for missing on some steps on "contribution protocol". I took your suggestion `const contactFormTypeLabel = i18n[contactFormType];`

christianbundy:
No apologies necessary, I appreciate the time you spent on this patch. Thanks!

Issue #361, opened 2020-03-26
Lock icon showing up above timestamp


christianbundy:
## What's the problem you want solved?

![Screenshot_2020-03-26 Oasis](https://user-images.githubusercontent.com/537700/77703911-24f54080-6f79-11ea-959a-c8d063ed1ca5.png)

## Is there a solution you'd like to recommend?

:man_shrugging: Probably some CSS baloney.

nickwynja:
This is probably related to my changes. I don't have any locked (private?) messages so never came across this. I'll take a look.

christianbundy:
Just sent you a private message!

christianbundy:
:muscle:

PR #362, opened 2020-03-27
Revert selector to span to fix lock alignment

nickwynja:

#361


christianbundy:
Super clever, this would've taken me a while. Thanks for this patch.

PR #339, opened 2020-03-23
Automatically connect to replicated peers

christianbundy:
Problem: By default we only connect to people that we're directly
following, which often don't have public IP addresses.

Solution: Automatically connect to all available peers within our
replication hops (default 2) to make peering more reliable.

PR #363, opened 2020-03-27
Add TypeScript definitions for better guardrails 

christianbundy:
Problem: We have TypeScript enabled on the project but have been missing
lots of definition files, so there are like 500 errors when you enable
`--noImplicitAny`.

Solution: Add definitions and fix the bugs that they point out. This
reduces the number of errors with `--noImplicitAny` to ~~285~~ 275.

---

This branch has types for unmerged pull requests (#339 and #341), so I'm waiting until someone has time to merge those before this PR is ready for review.


src/models.js
```diff
@@ -469,10 +469,17 @@ module.exports = ({ cooler, isPublic }) => {
...
+          lodash.get(msg, "value.content.text", false) !== false;
+        const hasRoot = lodash.get(msg, "value.content.root", false) !== false;
+        const hasFork = lodash.get(msg, "value.content.fork", false) !== false;
+
         const channel = lodash.get(msg, "value.content.channel");
         const hasChannel = typeof channel === "string" && channel.length > 2;
+        const isRoot = hasRoot === false;
 
-        if (hasChannel) {
+        if (hasChannel && isRoot) {
```
jedahan: does this change behaviour?

christianbundy: Yes, this bit from the PR description is relevant here:

> This branch has types for unmerged pull requests (#339 and #341), so I'm waiting until someone has time to merge those before this PR is ready for review.

This code is from #341, which makes it so that we only display the `channel` on root posts that set the channel, not on every reply to that post.

PR #364, opened 2020-03-27
Use blockquote on replies for text-based browser indent

nickwynja:

In a text-based browser, replies in threads were displayed at the same level, making it hard to distinguish what was the root. Relates to #329 and #336.

Text-based browsers recognize `<blockquote>`
and indent as expected. This helps render threads with indents in a
similar way as Oasis now does in the graphical browser.

This renders as:

```
[]Kathleen Marie 2h

Enjoying all the neighborhood flowers that are in bloom.

IMG_0186.jpg
IMG_0190.jpg

❤ 5
Comment JSON


    []Christian Bundy commented on thread 1h

    @Kathleen Marie

    Walks make lockdown so much more fun.

    ❤ 3
    Comment Reply JSON
```

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -534,8 +534,10 @@ hr {
  * that we don't need any inline CSS anymore.
  */
 .indent {
-  padding-left: 1rem;
+  padding-left: var(--whole);
   border-left: var(--micro) solid var(--bg-selection);
+  margin-right: 0;
+  margin-left: var(--whole);
 }
 
 .theme-preview {
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -8,6 +8,7 @@ const MarkdownIt = require("markdown-it");
 const {
   a,
   article,
+  blockquote,
   br,
   body,
   button,
@@ -422,7 +423,7 @@ const postAside = ({ key, value }) => {
     fragments.push(section(continueThreadComponent(thread, isComment)));
   }
 
-  return div({ class: "indent" }, fragments);
+  return blockquote({ class: "indent" }, fragments);
 };
 
 const post = ({ msg, aside = false }) => {


christianbundy:
Nice! Reading threads in text browsers is certainly a pain point, and I'd be happy to merge this. Two notes, just double-checking that this is intended.

On my screen it looks like the indentation is applied differently:

```
   Kathleen Marie 2h

   Enjoying all the neighborhood flowers that are in bloom.

   IMG_0186.jpg
   IMG_0190.jpg

   (0) M-$ 5
   Comment JSON

   Christian Bundy commented on thread 2h

     @Kathleen Marie

     Walks make lockdown so much more fun.

   (1) M-$ 3
   Comment Reply JSON
```

If it's useful:

```console
$ lynx --version
Lynx Version 2.8.9rel.1 (08 Jul 2018)
libwww-FM 2.14, SSL-MM 1.4.1, OpenSSL 1.1.1e, ncurses 6.2.20200212
Built on linux-gnu (Sep 10 2018 22:39:07).
```

It also seems to me that `<blockquote>` has a very specific use-case:

> The blockquote element represents a section that is quoted from another source.
>
> Content inside a blockquote must be quoted from another source, whose address, if it has one, may be cited in the cite attribute.
>
> -- https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element

I wouldn't lose any sleep worrying about what the standards committees will think, but I wonder whether the `<blockquote>` semantic is used by friends with accessibility needs. My hunch is that recursively nested blockquotes could be super confusing, but if you think otherwise I'm happy to trust your judgement and merge away.

nickwynja:
Hmm, interesting. I assumed that all text-based browsers would handle `<blockquote>` the same. I tested in `w3m` but that's all.

```
$ w3m -version
w3m version w3m/0.5.3+git20190105, options lang=en,m17n,color,ansi-color,mouse,menu,cookie,ssl,ssl-verify,external-uri-loader,w3mmailer,nntp,ipv6,alarm,mark
```

I think it's reasonable to close this since it doesn't really solve the problem in a cross-browser compatible way and introduces semantic/accessibility issues in the process.

christianbundy:
I was thinking about this earlier and I have a thought: maybe the reason the indentation isn't working is because we're using CSS variables for all the sizes and Lynx doesn't understand CSS variables? See also #375.

christianbundy:
(If that isn't the root problem, please open an issue so we can continue discussing and find a resolution!)

Issue #318, opened 2020-03-13
Chatting about Oasis


christianbundy:
## What's the problem you want solved?

There's no place to chat about Oasis. There are GitHub issues for tracking problems and SSB messages, but neither feel very lightweight for casual conversation.

## Is there a solution you'd like to recommend?

I've started a Matrix room so that people can stop by and chat:

https://riot.im/app/#/room/!YQpqIZlvBVPYRwrkXp:matrix.org

Maybe we could/should add it to the docs?

I was originally thinking that a Signal group would be nice, but it's not as easy to join and requires phone numbers. Is Matrix is a good compromise?

## What version or commit of Oasis are you using?

latest

aadilayub:
I think Matrix is a good choice!

PR #365, opened 2020-03-27
Add Matrix chat room to readme

christianbundy:
Problem: There's a chat room but it isn't in the readme.

Solution: Put it in the readme.

Fixes: https://github.com/fraction/oasis/issues/318


sbillig:
There's a chat room??

Issue #358, opened 2020-03-25
Don't show a post's channel twice under the post in the popular view


Powersource:
## What's the problem you want solved?

Posts posted *in* a channel (and not just with the channel mentioned) get that channel rendered twice in at least the popular view, like this

![image](https://user-images.githubusercontent.com/1427063/77594825-db184600-6ef7-11ea-935b-10c8c8a9d554.png)

(in some cases people both post in a channel, and tag the channel at the end of the post, but in that case it would look like it's rendered 3 times!)

## Is there a solution you'd like to recommend?

I'm guessing it's a bug, so fix it :P Note, this is unrelated to https://github.com/fraction/oasis/issues/290

christianbundy:
Yep, it looks to me like `transform()` is getting called twice on one message, but an initial skim of the code didn't make it obvious why this is happening.

PR #366, opened 2020-03-27
Avoid processing popular posts twice

christianbundy:
Problem: Messages are being fetched with `post.get()`, which runs the
`transform()` function to decorate them with Markdown/etc, but then
they're being passed through `transform()` a second time at the end of
the function. This is inefficient and applies side-effects (like adding
channels to the post) twice.

Solution: Remove the final `transform()` so that these posts are only
decorated once.

Fixes: https://github.com/fraction/oasis/issues/358

|:---|---:|---:|---:|---:|
| `curl http://old/public/popular/day` | 908.5 ± 59.2 | 851.9 | 998.5 | 1.42 ± 0.16 |
| `curl http://new/public/popular/day` | 641.2 ± 57.2 | 579.6 | 751.6 | 1.00 |

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -1065,7 +1065,7 @@ module.exports = ({ cooler, isPublic }) => {
                   if (err) {
                     reject(err);
                   } else {
-                    resolve(transform(ssb, collectedMessages, myFeedId));
+                    resolve(collectedMessages);
                   }
                 })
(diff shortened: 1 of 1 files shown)


sbillig:
Looks good. Thank you for your contribution!

Issue #370, opened 2020-03-28
Block button missing


christianbundy:
## What's the problem you want solved?

The follow message is available, but the block message is not.

## Is there a solution you'd like to recommend?

Re-add the block link to the profile.

christianbundy:
Maybe it never existed? I could've sworn this was a thing I'd already implemented, maybe I forgot to make a PR.

PR #372, opened 2020-03-28
Add block and unblock to author profile

christianbundy:
Problem: We have a way to follow and unfollow, but there's no way to
block or unblock. Also if you go to the profile of a blocked peer, their
posts still show up.

Solution: Add block and unblock and hide messages from blocked peers on
their profile page.

Fixes: https://github.com/fraction/oasis/issues/370

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -353,7 +353,7 @@ router
       name,
       description,
       avatarUrl,
-      relationship: null,
+      relationship: { me: true },
     });
   })
   .get("/profile/edit", async (ctx) => {
--- a/src/index.js
+++ b/src/index.js
@@ -681,6 +681,18 @@ router
     ctx.body = await friend.unfollow(feed);
     ctx.redirect(referer.href);
   })
+  .post("/block/:feed", koaBody(), async (ctx) => {
+    const { feed } = ctx.params;
+    const referer = new URL(ctx.request.header.referer);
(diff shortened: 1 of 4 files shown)


sbillig:
@christianbundy 

When viewing my own profile, I get:

```
TypeError: Cannot destructure property 'blocking' of '(intermediate value)' as it is null.
    at Object.fromPublicFeed (/Users/sean/code/oasis/src/models.js:558:15)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async author (/Users/sean/code/oasis/src/index.js:275:24)
    at async /Users/sean/code/oasis/src/index.js:289:16
    at async middleware (/Users/sean/code/oasis/src/index.js:802:7)
    at async /Users/sean/code/oasis/src/index.js:780:5
    at async /Users/sean/code/oasis/src/index.js:775:5
    at async /Users/sean/code/oasis/src/http.js:29:5
    at async /Users/sean/code/oasis/node_modules/koa-mount/index.js:52:26
```

christianbundy:
F.

Thanks for pointing that out, I'll push a fix. I should probably also add some basic sanity tests so we can *at least* send HTTP requests to all the basic URLs and make sure they don't throw errors. :man_facepalming:

PR #377, opened 2020-03-31
Show who liked content on hover

### LINKED ISSUE #256: How do I see who hearted something?
**What's the problem you want solved?** 

Someone hearted my post. I don't know who. I tried clicking on it to get a list, but it just liked my post instead.

Let me hover over the like button and get a list of people who like it. Or let me click and see the list.

timjrobinson:
- Adds a hover popup that shows the names of everyone who liked a post
when hovering of the heart.
- Add new call to post.get that retrieves the names of all voters and
returns them instead of their ID's.

I added the extra "like" class to make the css more specific, didn't want to apply the hover effect to all buttons. 

Fixes #256 

A simple hover on all hearts that shows who liked something. Here's how it looks for a post with a few likes:

2020-03-30-235113_468x288_scrot

And a post with many likes:

2020-03-30-235105_416x306_scrot

Happy to make changes to CSS to fit Oasis better, I'm not the best designer.

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -474,7 +474,7 @@ const post = ({ msg, aside = false }) => {
 
   const maxLikedNameLength = 16;
   const maxLikedNames = 20;
-  
+
   const likedBy = msg.value.meta.votes
     .slice(0, maxLikedNames)
     .map((name) => name.slice(0, maxLikedNameLength))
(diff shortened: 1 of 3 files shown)


src/views/index.js
```diff
@@ -467,11 +467,19 @@ const post = ({ msg, aside = false }) => {
...
 
   const likeButton = msg.value.meta.voted
-    ? { value: 0, class: "liked" }
-    : { value: 1, class: null };
+    ? { value: 0, class: "like liked" }
+    : { value: 1, class: "like" };
 
   const likeCount = msg.value.meta.votes.length;
 
+  const maxLikedNameLength = 16;
```
timjrobinson: Wasn't sure where to put settings like these, didn't want to leave them as magic numbers.

christianbundy:
This is great, thanks for putting this patch together. Merging now. :shipit:

PR #378, opened 2020-03-31
Updates query for threads

nickwynja:
Previous query incorrectly pulled and sorted posts, often showing years
old posts from newly followed people.

This now behaves more consistently
as "recent threads from people in your extended network".


christianbundy:
Nice catch! I thought I'd removed all those `messagesByType` methods but I'm happy you found this. :tada:

PR #379, opened 2020-03-31
Fix z-index on sticky header

nickwynja:
##  What's the problem you solved?
Screen Shot 2020-03-31 at 11 05 18 AM

Screen Shot 2020-03-31 at 11 05 26 AM


christianbundy:
Nice catch!

PR #367, opened 2020-03-27
Italian translation, changed a string in the Spanish translation

wuhei:
## What's the problem you solved?

Added Italian translation

## What solution are you recommending?

Double check the Spanish translation to see if there are discrepancies between Castillan and various versions of Spanish used in Southern and Central America.

PR #381, opened 2020-03-31
Fix linter problems with Italian translation

christianbundy:
Problem: #367 doesn't pass the linter.

Solution: Run `npm run fix` and disable the spell-checker when we
define language names.


christianbundy:
:bowing_man:

PR #382, opened 2020-03-31
Show liked-by in title attr, handle many likes

### LINKED ISSUE #380: List of people who have liked post requires mouse

The new code from #377 looks great when you're using a mouse on a browser like Chromium or Firefox, but it's inaccessible for people using other browsers (e.g. Lynx) or accessibility tools.

One elegant solution would be to use the [title attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) so that the browser can show this information natively. In Chromium/Firefox/etc this appears as a hover tooltip, but other user agents can display it however they'd like.

cc: @timjrobinson

timjrobinson:
- Instead of having a custom on hover effect, just add the Liked by
message to the title of the heart.
- When there are > 16 likes on a post, show +X more to convey this to
users.
- Removed the old css code and the 'like' class on the button, reverting CSS changes in #377 

Fixed #380 

Making the liked-by popup more accessible to non-mouse users. Also better for future maintainence as it doesn't require additional CSS or classes. 

Showing liked-by via title attribute, as described in #380

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -411,25 +411,6 @@ section > footer > div > form > button.liked {
   color: var(--red);
 }
 
-section > footer > div > form > button.like > span.liked-by {
-  display: none;
-}
-
-section > footer > div > form > button.like:hover {
-  background-color: inherit;
-}
-
-section > footer > div > form > button.like:hover > span.liked-by {
-  display: block;
-  position: absolute;
-  max-width: 300px;
-  padding: 15px;
-  background-color: var(--base07);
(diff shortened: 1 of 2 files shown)


christianbundy:
This looks great! The only bug I've found is that if nobody has liked a post, then it just shows "Liked by". Do you think that's something you have time/energy/interest in taking care of in this PR? If not, I'm happy to take care of it over the next few days.

christianbundy:
Looks great, thanks for that fixup! Merging now.

PR #388, opened 2020-04-03
Fix config bug

christianbundy:
Problem: I made a bad merge in 2836c80 which broke the server in some
environments. I had issues running the globablly installed binary,
whereas `npm start` seemed to work fine. Anyway, there's an error about
some missing config options because we weren't importing SSB-Config into
the server configuration. Instead, the full config was just:

```json
{ "conn": { "autostart": true } }
```

This lacks important properties like `shs`, which means that the server
can't start. Fun!

Solution: Pass SSB-Config first and then overlay our custom config on
top. I've also added comments and changed the variable names so that
this is harder to miss in the future.

### DIFF
--- a/src/ssb.js
+++ b/src/ssb.js
@@ -57,7 +57,7 @@ const rawConnect = (options) =>
 
 let handle;
 
-const createConnection = (config) => {
+const createConnection = (customConfig) => {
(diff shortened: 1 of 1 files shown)


christianbundy:
Thanks for the merge!

PR #389, opened 2020-04-03
Add very basic test suite

christianbundy:
Problem: ESLint and TypeScript help catch some types of regressions, but
they don't protect us against obvious stuff like "the server won't
start". This means that humans need to test a bunch of stuff manually,
and that can be really tedious and exhausting.

Solution: Yesterday someone invented this cool concept called "testing"
where you write automated tests for your software to ensure it actually
works the way you expect. It might have beeen invented before yesterday,
I don't know. Anyway, this solution adds a bunch of tests that send HTTP
GET requests to a bunch of endpoints to make sure the server is at least
returning HTTP 200 responses. It also fixes a race condition where HTTP
server was available before the readme / version strings were loaded.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+.nyc_output
--- a/package-lock.json
+++ b/package-lock.json
@@ -389,6 +389,12 @@
       "integrity": "sha1-GMyw/RJoCOSYqCDLgUnAByorkGY=",
       "dev": true
     },
+    "@types/cookiejar": {
+      "version": "2.1.1",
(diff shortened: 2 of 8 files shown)


christianbundy:
Once this is merged I think this will be the SSB client with the most test coverage. :scream:

cinnamon-bun:
⭐️ ⭐️ ⭐️ ⭐️ ⭐️  Yes!!

It even reports code coverage, nice!

christianbundy:
Thanks for noticing! :taco: I'm actually weirdly excited about the test coverage, that'll be a fun metric to play around with.

PR #393, opened 2020-04-09
Fix private posts rendered alongside public posts

christianbundy:
Problem: The Threads view advertises that it only shows public posts,
but is showing some private posts as well. This is not a security
concern, it's just a UI inconsistency.

Solution: Ensure that private posts are filtered from the view before
rendering. This also integrates a refactor that uses functions to check
for basic info like whether a message is a post, whether it's
encrypted/decrypted, whether it has a root, whether it has a fork, etc.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -10,6 +10,29 @@ const pullSort = require("pull-sort");
 const ssbRef = require("ssb-ref");
 const crypto = require("crypto");
 
+const isEncrypted = (message) => typeof message.value.content === "string";
+const isNotEncrypted = (message) => isEncrypted(message) === false;
+
+const isDecrypted = (message) =>
+  lodash.get(message, "value.meta.private", false);
+
+const isPrivate = (message) => isEncrypted(message) || isDecrypted(message);
+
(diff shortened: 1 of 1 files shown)


src/models.js
```diff
@@ -300,36 +321,24 @@ module.exports = ({ cooler, isPublic }) => {
...
   const isLooseRoot = (message) => {
     const conditions = [
       isPost(message),
-      lodash.get(message, "value.content.root") == null,
-      lodash.get(message, "value.content.fork") == null,
+      hasNoRoot(message),
+      hasNoFork(message),
     ];
 
     return conditions.every((x) => x);
```
cinnamon-bun, line 331: Is there a reason to use `conditions.every` here, instead of something like:

`return isPost(message) && hasNoRoot(message) && hasNoFork(message)`

christianbundy: Nope, I think I was trying to be clever and/or assuming that there would be tons of conditions.

christianbundy:
Thank you! It's so easy to get all tangled up in a bunch of comparisons and not know whether they're checking the same thing. :sweat_smile:

PR #394, opened 2020-04-10
Added french "fr" translation

zicmama:
FRENCH

Add fr language

### DIFF
--- a/src/views/i18n.js
+++ b/src/views/i18n.js
@@ -693,6 +693,184 @@ const i18n = {
     hashtagDescription:
       "Post da persone nella tua rete che menzionano questo hashtag, ordinati cronologicamente.",
   },
+    fr: {
+    // navbar items
+    extended: "Étendue",
+    extendedDescription: [
+      "Messages de ",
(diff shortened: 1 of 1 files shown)


georgeowell:
@zicmama Thanks for this patch. Busy time at the moment. I'm sure one of the maintainers will review this when they get the chance.

christianbundy:
Thanks for this patch! It looks like the tests are failing, do you need help with that?

There's some information in the [contributing guide](https://github.com/fraction/oasis/blob/master/docs/contributing.md), but the gist is that if you run `npm test` it should report what the problem is. There's a [Frequently Failed Tests](https://github.com/fraction/oasis/blob/master/docs/contributing.md#frequently-failed-tests) section that enumerates the usual suspects, but if you don't find your problem there I'd be happy to help you debug it.

I think you just need to run `npm run fix` and push a commit with the fix, but if it gives you any trouble I'm happy to help.

zicmama:
In fact, I didn't use npm, just "edit" the file by hand in github web interface and added a fr section to the file as other language have done... Maybe coma or bracket is missing...

PR #395, opened 2020-04-10
Start SSB service immediately instead of waiting

christianbundy:
Problem: During a refactor the SSB connection management was changed to
be more conservative, so it only ensures that we have a connection once
the server is started. This isn't good, because it means `oasis
--no-open` no longer starts an SSB service in the background.

Solution: Run `ensureConnection()` to ensure that we have a connection
to the SSB service regardless of regardless of whether we've received
any requests over HTTP.

---

See: `%feSvhuLLa8vA475+k1Z9ZWEAHvNhYTT1H4wlpcoGEYw=.sha256`

### DIFF
--- a/src/ssb.js
+++ b/src/ssb.js
@@ -61,6 +61,7 @@ const connect = (options) =>
 
 let closing = false;
 let serverHandle;
+let clientHandle;
 
 /**
  * Attempts connection over Unix socket, falling back to TCP socket if that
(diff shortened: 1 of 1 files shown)


christianbundy:
I FOUGHT THE BUILD AND THE BUILD WON.

But then I kept fighting the build, and eventually I won.

[Persistence hunting](https://en.wikipedia.org/wiki/Persistence_hunting) confirmed.

christianbundy:
Thanks for the merge!

PR #397, opened 2020-04-11
Fix 'liked by' hover in replies in Thread view

sbillig:

The 'liked by' title (hover) wasn't present in replies in the Threads view; now it is.

post() and postInAside() were basically the same, apart from the missing 'liked by' title in the latter, and an additional "post" css class in the former, so I removed postInAside() and used post() in its place.

The "post" class doesn't seem to do anything, so I left it in for the replies (which now have class = "post reply", instead of just "reply"). Lemme know if I should change this.

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -144,134 +144,6 @@ const template = (...elements) => {
   return result;
 };
 
-const postInAside = (msg) => {
-  const encoded = {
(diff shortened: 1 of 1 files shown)


christianbundy:
Deleting lines *and* fixing a bug? Almost too good to be true. Thanks for fixing this!

PR #398, opened 2020-04-11
Add --allow-host CLI flag for for proxy use-case

christianbundy:
Problem: When proxying Oasis, the CSRF + DNS rebind security precautions
will respond with HTTP 400 if you use a hostname that Oasis doesn't
know about. For example, if Oasis is listening on `localhost` and you
use Caddy/Nginx/etc to proxy that to `oasis.example.com`, then Oasis
will see GET requests as DNS rebind attacks and POST requests as CSRF
attacks.

Solution: Add `--allow-host` command-line flag so that you can
`--allow-host oasis.example.com` and ensure that the host is allowed by
the security measures.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -31,6 +31,12 @@ module.exports = (presets, defaultConfigFile) =>
       default: _.get(presets, "host", "localhost"),
       type: "string",
     })
+    .options("allow-host", {
+      describe:
+        "Extra hostname to be whitelisted (useful when running behind a proxy)",
+      default: _.get(presets, "host", null),
+      type: "string",
+    })
     .options("port", {
       describe: "Port for web app to listen on",
       default: _.get(presets, "port", 3000),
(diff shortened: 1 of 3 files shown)


sbillig:
Code looks reasonable and @seekr says this works, which is good enough for me!

pokapow:
Is it possible to alloaw a subpage, not just a subdomain?

`--allow-host example.com/oasis`

Actualy I get this error in console, without subpage in allow-host:

`Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-4Su6mBWzEIFnH4pAGMOuaeBrstwJN4Z3pq/s1Kn4/KQ='), or a nonce ('nonce-...') is required to enable inline execution.`

christianbundy:
It's not, unfortunately. I'd strongly recommend running this as a subdomain rather than as a sub-path.

pokapow:
I understand, but we use local hostnames of machines, /etc/hosts is not default configure with subdomain of the main local domain ...

I think that should be a nice trick to allow this usecase.

Issue #401, opened 2020-04-13
`setImmediate()` doesn't always work.


christianbundy:
## What's the problem you want solved?

It's been reported on Windows that after `setImmediate()` in `http.js` the `server` variable is still `null`. I can't reproduce this, but it seems reasonable. `setImmediate()` is an awful little hack.

## Is there a solution you'd like to recommend?

I think maybe we need `server.on('listening')` or something instead?

PR #402, opened 2020-04-13
Replace setImmediate() with .on('listening')

christianbundy:
Problem: Unfortunately `setImmediate()` doesn't actually ensure that
`server !== null`, so we can get fun race condition errors that have
been reported on Windows.

Solution: Replace `setImmediate()` hack with event emitter that actually
ensures that the server is listening (and hopefully isn't `null)`.

Fixes: https://github.com/fraction/oasis/issues/401


christianbundy:
Woop, thank!

Issue #354, opened 2020-03-25
Add command to do complete sync


nickwynja:
## What's the problem you want solved?

As mentioned in `%25b2I4vI7sRFKB5N9iNVv54aNZLWFOI6ioSQ8cKiv0%2B8k%3D.sha256`, I'm looking for a command like `oasis --sync` that can be run to do the following:

1. Start a server or connect to an exist `ssb-server`
1. Start a connection to peers if you aren't already
1. Sync the feed
1. Get `blob.wants`
1. Return the `conn` state to how it was, if previously stopped
1. Kill `ssb-server` if it started one
1. Successfully exit the script with all the things downloaded

I've created [this shell script](https://gist.github.com/nickwynja/721f98b50941fa618c99e9c40548db37) that does much of what I'm looking to do in a pretty messy way. It also repeats much of the core functionality that is done much better in Oasis.

The use case is something like this: someone lives on a boat without any data connection. They come into dock and quickly get a coffee from a cafe before having to leave to do other boat things. They can open their laptop and run `oasis --sync` or have a script that runs `oasis --sync` among several other sync tasks like email before shutting down their machine and getting back to the boat. By having executed the `blob.wants` during the sync script, when this person is back on their boat they are able to have a fully-synced (as of that point in time) offline experience, without missing images or other assets.

## Is there a solution you'd like to recommend?

Something that helps to get a full offline experience.

nickwynja:
Another option could be to have getting `blob.wants` managed as a separate option so, for #328, you could do `oasis --sync --no-blobs`.

christianbundy:
Another way that might be a bit easier to implement: when we're offline, show a "sync" button that:

- Goes online
- Waits for new messages to appear
- Waits for the message count to stop increasing for ~30 seconds
- Goes back offline

christianbundy:
Once we have that, it would be trivial to add a `--sync` command-line option that starts a new server and makes the first action a sync. The only bit that's important to me is that I'd love to avoid needing to turn off the server just to trigger another sync.

aadilayub:
> Another way that might be a bit easier to implement: when we're offline, show a "sync" button that:
> 
> * Goes online
> * Waits for new messages to appear
> * Waits for the message count to stop increasing for ~30 seconds
> * Goes back offline

[...]

PR #403, opened 2020-04-13
Introduces sync endpoint

nickwynja:

Constant network activity of connecting to peers and streaming of data can be really busy and distracting. "Sync" allows you to keep Oasis running in offline mode and sync when you want to get new stuff. Sync can be expanded upon in the future to be more useful when calling from a script.

Relates to #354. 

When clicking "Connect and Sync" in settings, Oasis starts the connection scheduler, waits until you're connected to peers, waits for items to download, and then disconnects from peers.

This is best used when running Oasis in offline mode. It's basically useless in "normal" mode and will actually stop the connection scheduler and disconnect you from connected peers, which may be unexpected if you're not already offline. This could probably be improved upon.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -342,6 +342,7 @@ module.exports = ({ cooler, isPublic }) => {
       // Let's wait a second for things to be ready
       await delay(1000);
       const originalProgress = await ssb.progress();
+      const originalTarget = originalProgress.indexes.target;
 
       const waitForPeers = async () => {
         const peers = await models.meta.connectedPeers();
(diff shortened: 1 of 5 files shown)


nickwynja:
I think this still needs a bit of work, actually. I'll pull in the updated service PRs when they're merged and have to add some checks to make sure I'm waiting enough time for progress to update.

christianbundy:
I'm very excited about this btw. I added some notes in the Matrix chat, probably should have done so here instead. Let me know if you'd like to pair on this sometime!

src/models.js
```diff
@@ -287,6 +286,18 @@ module.exports = ({ cooler, isPublic }) => {
...
+
+      if (peersList !== undefined) {
+        peersList.forEach(([address, data]) => {
+          // debug(address, data.state);
+          if (data.state === "connected") {
+            peersConnected.push([address, data]);
+          }
+        });
+      }
+      return peersConnected;
```
nickwynja, line 300: This seems really nasty and would love any recommendations on how to tighten this up.

nickwynja:
@christianbundy I think this is in better shape now. It would be very much appreciated if you could review the progress check logic. I don't think the nested loop I'm doing is the best way possible but I'm not sure of another clean way to do it and keep track of some state. Thanks!

christianbundy:
This looks really good, I was about to merge but wanted to check in on `indexes.current` vs `indexes.target`. Can we use 'target' everywhere? The 'target' value refers to how many bytes the database has downloaded, whereas 'current' refers to how many bytes we've indexed for the various views. Since this code only cares about downloads, I think we just want to use 'target' here, but you may be using it for a different reason?

Otherwise, this LGTM. :+1:

nickwynja:
> The 'target' value refers to how many bytes the database has downloaded, whereas 'current' refers to how many bytes we've indexed for the various views.

Ah, I see. I noticed in some cases that `target > current` and I was assuming that meant "I know `target` number of objects exist but I have only downloaded `current`".

I'll rework a bit to only use `target` and test some more.

christianbundy:
Sounds good, happy to merge after that! The only other [silly] thing is that the index integer refers to how many bytes are in the database rather than the number of objects. (Please disregard if this is too much information.) It's weird, but it means that you can read 4 bytes from `~/.ssb/flume/log.offset` offset N, parse those bytes as an integer (L), and then read from `N + 4` to `N + 4 + L` to get the message from the database. 

So if you have `indexes.target === 1000`, then you read from bytes 100 to 1004 and get an integer like 200, and then you read from bytes 1004 to 1204 to get the message from the database. Kinda janky, but also kinda fast. :shrug: 

More info here, if you're curious: https://github.com/flumedb/flumelog-offset#format

nickwynja:
Interesting. I never gave it more than a passing thought but was a bit surprised that think that there was nearly a billion objects... bytes makes much more sense. 😂

nickwynja:
Also worth mentioning that in absence of a CLI for `sync`, I've been using this `curl` command:
```
curl -f -X POST http://localhost:3000/settings/conn/sync --referer http://localhost:3000
```

PR #405, opened 2020-04-13
Add Français (FR) to the language drop-down

christianbundy:
Problem: The FR translation is added in `i18n.js` but it's not available
in the drop-down language selection menu.

Solution: Add it to the menu and ensure that the menu is sorted
alphabetically by the name of the language.

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -820,7 +820,7 @@ exports.settingsView = ({ status, peers, theme, themeNames, version }) => {
     })
   );
 
-  const languageOption = (shortName, longName) =>
+  const languageOption = (longName, shortName) =>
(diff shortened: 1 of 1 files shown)


src/views/index.js
```diff
@@ -820,7 +820,7 @@ exports.settingsView = ({ status, peers, theme, themeNames, version }) => {
     })
   );
 
-  const languageOption = (shortName, longName) =>
```
jedahan, line 823: how do you feel about `({longName, shortName})` to help reduce ordering issues?

christianbundy: Yep, named parameters via object destructuring are my preference. I'd merge a patch to change that. :+1:

PR #406, opened 2020-04-13
Change offline warning message

cinnamon-bun:
The WARNING message about offline mode talks about "pre-existing servers", meaning other SSB apps on your own computer.  I wouldn't expect people to understand that.

Rewrite warning in less technical language.


christianbundy:
This is so good, thanks for fixing it. I didn't realize how bad the previous message was until I saw yours!

PR #409, opened 2020-04-14
Add timeout to sync and refactor

christianbundy:
Problem: The sync goes forever, which isn't good if we're in an edge
case (like having peers who are offline with us).

Solution: Add a 5-minute timeout and refactor to use a design that's
maybe simpler? This is just my stab at the previous behavior, but there
are definitely improvements that could be made!

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -327,43 +327,44 @@ module.exports = ({ cooler, isPublic }) => {
       const ssb = await cooler.open();
 
       const progress = await ssb.progress();
-      let previousTarget = progress.indexes.target
+      let previousTarget = progress.indexes.target;
 
       // Automatically timeout after 5 minutes.
       let keepGoing = true;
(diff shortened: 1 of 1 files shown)


christianbundy:
Sample output:

```
  oasis Starting sync, waiting for new messages... +0ms
  oasis Difference: 0 bytes +5s
  oasis Received no new messages. +0ms
  oasis Difference: 0 bytes +5s
  oasis Received no new messages. +0ms
  oasis Difference: 0 bytes +5s
  oasis Received no new messages. +1ms
  oasis Difference: 0 bytes +5s
  oasis Received no new messages. +0ms
  oasis Difference: 541 bytes +5s
  oasis Finished waiting for first new message. +0ms
  oasis Difference: 0 bytes +5s
  oasis Finished waiting for last new message. +0ms
```

PR #410, opened 2020-04-14
Fix summary style

christianbundy:
Problem: the `<summary>` element had a broken bottom margin on content
warnings, and isn't obvious that it's not just text.

Solution: Fix the margin and used a dashed border to indicate that the
content can be expanded. I think this is more intuitive than no border,
and I think the dashed line draws parallels with "cut outs" that might
make expansion intuitive (?), but either way I think it's better than
the previous design.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -448,11 +448,12 @@ progress {
 
 summary {
   padding: var(--milli);
-  margin-top: var(--whole);
+  margin: var(--whole) 0;
   cursor: pointer;
   background: var(--bg);
   border-radius: var(--common-radius);
   list-style-type: "+ ";
+  border: var(--pico) dashed var(--fg-status);
 }
 
 details[open] > summary {
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -445,7 +445,7 @@ progress {
 
 summary {
   padding: var(--size--1);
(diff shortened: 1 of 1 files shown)


christianbundy:
cc: @nickwynja, I'd be curious to hear what you think about this.

nickwynja:
I think this works nicely for content warning! It doesn't feel quite right for reply summaries though. It makes the reply jump out visually from the thread when the intent of the summary is to soften the effect of the reply in the overall thread. I'm not sure it would make sense to remove the border for `.indent > details > summary` though since we wouldn't be using the visual style consistently. Hmm.

christianbundy:
Alternative: Is this too much contrast?

```css
section {
  border: var(--pico) solid var(--fg-status);
}
```

I recently found https://www.stayhometakecare.com/ and I'm kinda inspired by the aesthetic.

nickwynja:
> Is this too much contrast?

It's maybe a little strong but that's subjective (and possibly dependent on which theme someone is using).

christianbundy:
I don't mind removing the border on ` .indent > details > summary` if you prefer, just let me know. The important thing to me is that this patch doesn't cause more problems than it solves.

nickwynja:
Yea, maybe it's most useful just on the content warning `<summary>` at this point.

Issue #100, opened 2020-01-28
Metric CSS is a pain point


christianbundy:
**What's the problem you want to solved?** I thought the metric-derived CSS variables were cute and fun, but I tried to edit the CSS today and I couldn't remember whether to use `--micro` or `--milli` for half-size. Boo.

**Is there a solution you'd like to recommend?** Maybe we should just do:

```css
:root {
  --three: calc(var(--one) * 3);
  --two: calc(var(--one) * 2);
  --one: 1rem;
  --half: calc(var(--base) / 2);
  --third: calc(var(--base) / 3);
  --quarter: calc(var(--base) / 4);
}
```

This has the nice side-effect of being able to easily change the base, so we could use 1.5rem or the golden ratio or whatever we want in the future without having to memorize quirky variables names. Simplicity!!!!!

Powersource:
A convention I kind of like is to call it e.g. `--size-1` `--size-2` etc, makes it easy to switch between different ones and I think they're sorted when autocomplete in devtools suggests them. (:bike::house:)

georgeowell:
This is quite stale. Did you decide what to do yet?

christianbundy:
It's old, but it's still a problem that I'd like to resolve.

christianbundy:
cc: @nickwynja if you have any thoughts

PR #411, opened 2020-04-14
Replace "metric CSS" with --size-N

christianbundy:
Problem: The "metric CSS" was kind of clever but mostly just confusing.

Solution: Use `--size-N` for the size, where the size is `2^n`. Double
size is `--size-1`, half size is `--size--1` (negative one). The
negatives kind of suck but I couldn't think of a better solution that
didn't give up the flexibility, brevity, or the ability to sort
alphabetically.

Fixes: https://github.com/fraction/oasis/issues/100

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -29,22 +29,19 @@
   --fg-light: var(--base06);
   --bg-light: var(--base07);
 
-  /* line */
-
-  --peta: 32rem;
-  --tera: 16rem;
-  --giga: 8rem;
-  --mega: 4rem;
-  --kilo: 2rem;
-  --whole: 1rem;
-  --milli: 0.5rem;
-  --micro: 0.25rem;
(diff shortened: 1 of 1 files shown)


src/assets/style.css
```diff
@@ -465,7 +462,7 @@ details[open] > summary {
   -webkit-user-select: all;
   user-select: all;
   background: none;
-  font-size: 50%;
+  overflow: hidden;
```
christianbundy, line 465: I've also [accidentally] snuck in an improvement to the Markdown mention on the profile page, replacing the itsy-bitsy font with `overflow: hidden`.

cinnamon-bun:
This is an improvement!  The metric sizes were a bit confusing to work with.

`--size--1` looks very similar to `--size-1`... would `--size-n1` work for negative numbers?

PR #413, opened 2020-04-15
Add super simple same-as implementation

christianbundy:
Problem: Most of us have lots of feeds and there's no way to indicate
which feeds we want to see notifications for.

Solution: Start reading the `sameAs` profile property, and if it's an
array with feed IDs then also look up mentions for those feed IDs. Note
that this commit doesn't introduce a way to add same-as in the UI, so
you ~~have~~ get to use the custom message publish feature to make a
message like:

```json
{
  "type": "about",
  "about": "<your current feed id>",
  "sameAs": [
    "<another feed id>"
  ]
}
```

The `sameAs` array can have as many items as you'd like, although I've
only actually tested with two.

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -157,12 +157,37 @@ module.exports = ({ cooler, isPublic }) => {
       return raw;
     },
     sameAs: async (feedId) => {
-      const raw =
-        (await getAbout({
-          key: "sameAs",
-          feedId,
-        })) || [];
-      return raw;
+      const getSameAs = async (someFeedId) => {
+        const feedList = await getAbout({ key: "sameAs", feedId: someFeedId });
+        if (Array.isArray(feedList)) {
+          return feedList;
+        } else {
+          return [];
+        }
+      };
+
+      // Get list of assertions.
+      const asserted = await getSameAs(feedId);
+
+      // Check assertions from all fo those feeds and filter by whether they include us.
+      const results = await Promise.all(
+        asserted.map(async (target) => {
+          const targetSameAs = await getSameAs(target);
+
+          if (targetSameAs.includes(feedId)) {
+            return target;
+          } else {
+            return null;
+          }
+        })
+      );
+
+      // Remove feeds that didn't assert same-as for our feed ID.
+      const confirmed = results.filter((target) => target !== null);
+
+      debug({ asserted, confirmed });
+
+      return confirmed;
     },
   };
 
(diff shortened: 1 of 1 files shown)


christianbundy:
I'll mark this 'ready for review' once I've got some feedback on SSB. This works great for the "I need to check my mentions for another feed" use-case, but it's important to me that the rest of the ecosystem can weigh in on this.

See: `%yiZqaIIO7puICdaqrAvuRyon+78ZhvrmGW7mnVjhOAs=.sha256`

cblgh:
to add consent without needing to implement stuff in other clients, i wonder if the following is a decent compromise:
* select the ssb identity `ident` you want to add to yr sameAs array
* create a regular post  (`type: 'post'`?), with text saying "i added you to sameAs!" and mentioning `ident` in the text
* create the sameAs message `type: 'sameAs'`, which _references the regular post_ and with `ident`as target
* IF the referenced post has been liked (`type: 'vote'`) by the targeted ssb identity `ident`, THEN consent has been given

feel free to swap out the `type: 'post'` for `type: 'about'` or whatever makes sense in actual ssb implementation land

christianbundy:
Weird timing! I spoke about this a bit in `%dEVWfUFJby4rRiCPUPXIj98sEwLJNmGUoK1rD9MchVw=.sha256` but I can reiterate a bit here if that's helpful?

Consent is important, but I'm looking for more information to understand how it applies for this feature.

Right now anyone can look up your mentions with SSB-Backlinks or even the search function, and I don't think I understand why we should add more usage steps here. It feels like we're in a house with two wide-open doors, but when we build a third door you want to make sure that it has a sophisticated lock on it.

My concern: there are already two open doors, and if there are bad actors then they can just remove the locks. I'm very hesitant to add the illusion of privacy around mentions, and to be honest I'm *also*  concerned about making SSB any harder to use. If we were announcing "these two people are the same" then **of course** we need consent from both parties, but "I want to see mentions for this other feed" doesn't feel to me like it needs explicit consent. I'd love to change my mind if you feel strongly, I think I just maybe need more info to understand your perspective more?

Powersource:
I think cblgh's protocol is maybe a bit complex (just mutual sameAs'ing would work fine even if there's no ui for that yet) but I agree that consent is needed. There's lots of creepy/weird stuff that a programmer can do with ssb, e.g. social graph analysis. That doesn't mean it's fine to add as a feature in a client just because it's technically possible.

christianbundy:
Oops! I amended my commit but forgot to `git add` first. This should be ready for review now.

cblgh:
for the sake of history: i wrote a reply here: %3xycnGpK1QDrxFjhzZAnifXawG2xHyubR0xOmqFpPvE=.sha256

Issue #27, opened 2019-12-16
Rename 'reply', 'comment', or both


christianbundy:
**What version of this package are you using?** N/A

**What problem do you want to solve?** "reply" and "comment" aren't great verbs

**What do you think is the correct solution to this problem?** We should give them good names! Is "fork" better? I'm struggling with how to express the difference between "add a post to extend the current thread" and "add a post as a tangent off of a message".

**Are you willing to submit a pull request to implement this change?** Yes!

christianbundy:
Maybe "sub-thread"?

jedahan:
I'm generally confused about threads. In the ux, I see "...replied to a forked thread" and I'm not sure what that means. Its some weird mix between twitter and zulip but somehow worse than both. I wonder if spending some time thinking about the UX we would love to see would help us with better verbs.

christianbundy:
That "replied to a forked thread" is coming from Patchwork, right? In Oasis there are three actions:

- **Root:** Start a new thread by creating a message without any ancestors.
- **Comment:** Add a response to an entire thread (which is a **root** or **reply**, plus its comments).
- **Reply:** Add a response to a specific **comment**.

I think Patchwork uses the word 'fork' instead of 'reply', but there's also this other unrelated thing where you make a new root and reference a message which people *also* call a 'fork'.

Also worth noting: Patchwork doesn't support adding a **comment** to a **reply**, they just add another reply as a sibling. There's a good discussion about all of this here: `%d6YLxDcLgHvhpoZx2K5uZig6ht8lTAtbACKFRty7q6Y=.sha256`.

![image](https://user-images.githubusercontent.com/537700/71919848-70e2d700-313a-11ea-8d10-f0915138e18e.png)

In this case:

- **A:** root
- **B, C, D, E:** comments on **A**
- **m, x**: replies to **C**
- **n:** comment on **m**
- **y:** comment on **x**

christianbundy:
Starting to lean toward 'tangent' instead of 'reply' or 'fork'. :thinking:

georgeowell:
Ah this is so hard! 

I'm kind of really into "Tangent"

We naturally tend towards "Fork" because of our interest in software but Tangent seems to describe that action so much better.

## Tangent Definitions
> Diverging from an original purpose or course.

> Digressing suddenly from one course of action or thought and turning to another.

> Origin of tangent
1585–90; < Latin tangent- (stem of tangēns, present participle of tangere to touch) in phrase līnea tangēns touching line.

[...]

PR #414, opened 2020-04-16
Rename "reply" to "subtopic"

christianbundy:
Problem: The words "comment" and "reply" are super ambiguous, but
renaming them is hard.

Solution: I think "subtopic" captures the sentiment well, and makes it
clear that you're leaving the current topic to start a new topic (rather
than just replying to an individual message).

Fixes: #27

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const debug = require("debug")("oasis");
-const { isRoot, isReply } = require("ssb-thread-schema");
+const { isRoot, isReply: isComment } = require("ssb-thread-schema");
 const lodash = require("lodash");
 const prettyMs = require("pretty-ms");
 const pullParallelMap = require("pull-paramap");
@@ -34,7 +34,7 @@ const isPost = (message) =>
   typeof lodash.get(message, "value.content.text") === "string";
 
 // HACK: https://github.com/ssbc/ssb-thread-schema/issues/4
-const isNestedReply = require("ssb-thread-schema/post/nested-reply/validator");
+const isSubtopic = require("ssb-thread-schema/post/nested-reply/validator");
 
 const nullImage = `&${"0".repeat(43)}=.sha256`;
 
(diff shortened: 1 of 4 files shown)


src/models.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 const debug = require("debug")("oasis");
-const { isRoot, isReply } = require("ssb-thread-schema");
+const { isRoot, isReply: isComment } = require("ssb-thread-schema");
```
cinnamon-bun, line 4: I find Object destructuring assignments to be counterintuitive so I had to look this up (I always think they work the opposite way around).

In this case we're taking `isReply` from `ssb-thread-schema` and assigning it to a local variable `isComment`.

Maybe a comment could help others avoid my confusion
`// Rename ssb-thread-schema's isReply to isComment

src/models.js
```diff
@@ -386,7 +386,7 @@ module.exports = ({ cooler, isPublic }) => {
     return conditions.every((x) => x);
   };
 
-  const isLooseReply = (message) => {
+  const isLooseSubtopic = (message) => {
```
cinnamon-bun, line 389: Could be useful to add comments explaining what "loose" messages are / under what circumstances they occur.  I see later in this file around line 1230 these functions are used to tell apart "thread replies" from "message replies" but I'm not sure what the difference is

src/models.js
```diff
@@ -698,7 +698,7 @@ module.exports = ({ cooler, isPublic }) => {
 
       return messages;
     },
-    threadReplies: async (rootId, customOptions = {}) => {
+    topicComments: async (rootId, customOptions = {}) => {
```
cinnamon-bun, line 701: Is a "thread" the same thing as a "topic"?  Maybe there's two words used to mean the same thing in various places

It might be useful to have a "vocabulary.md" document defining these words to help us use them consistently

christianbundy:
Thanks for the comments and the merge! I wish annotations just went directly into the source code, it sucks that they're only visible in this PR.

PR #425, opened 2020-05-03
Image search

Powersource:

I added a page for searching for images, like patchbay has :D 

image

---

Possible downsides (I don't think they're that bad tho):

So there's another sidebar item now which you maybe don't want. But I think this could maybe with not too much work be integrated into the regular search page (in another PR), removing this sidebar entry again.

There's just a few bugs, like it tries to render things that aren't images as well. But I think it works good-enough-for-now now and could be polished up more in the future.

And it needs another plugin for the server, `ssb-meme`. I'll submit the PR for that in a minute.

---

todo:
* [x] Move strings to i18n file
* [x] Merge that flotilla PR https://github.com/fraction/flotilla/pull/1
* [x] Make oasis use the latest version of flotilla


Powersource:
Here's the ssb-meme PR, gonna have to upgrade oasis to flotilla 5.0.1 first though https://github.com/fraction/flotilla/pull/1

Powersource:
ohh yeah just remembered that I want to add embed code just underneath the pictures so you can easily embed them. but can add that in another PR

christianbundy:
Nice! I just opened https://github.com/fraction/oasis/pull/427 which should upgrade us to the latest Flotilla and help with merge conflicts.

cinnamon-bun:
Awesome!

> I want to add embed code just underneath the pictures so you can easily embed them. but can add that in another PR

There are some SSB ecosystem impacts that could be mitigated with some small changes:

People can feel bad if images from their posts get reused elsewhere without attribution or understanding of the original context.  Examples:
* Art reposted without attribution
* A famous person reposting someone's personal photos to a large audience when they were originally intended for a smaller audience, etc.

Mitigations:
* Each image in the grid could link back to the first post that used it?  Can we do that query?
* Above the image grid, add some text "Please look at the original post to see if the author is OK with reposting their image"
* When copy-pasting the blob address, also include an attribution markdown link that will end up getting pasted into the new post.  The user might choose to remove that link before publishing, but it would set a cultural norm around attribution.

Personally I feel mixed about image search - it's really great for Totoro gifs and memes which make the space feel more colorful and welcoming.  But it also can let my photos go viral to a much larger audience than I thought I had, beyond the people who are following me, without my knowledge or consent.  That makes SSB feel less safe and cozy.

Open source culture is all about remixing stuff, but other cultures have strong feelings about reposting other peoples' content for reasons of attribution, privacy, or to maintain a "feeling of control" over their work and the context it's used in.

Anyway let's try it and see what happens?  I'm not aware of this problem happening yet from patchbay's image search though it has been an issue with content from the regular web being reposted into SSB (see `%cTcw+4MRs1xkGBok2mg35m/AR9z6mxzSxBx18s2RICo=.sha256`)

cinnamon-bun:
How does this interact with blobs in private messages?

Powersource:
> Each image in the grid could link back to the first post that used it? Can we do that query?

Due to the way blobs work it's not provable who uploaded the blob initially, and without some major causality math (using links between feeds to figure out who posted when) I don't think we could even make a very good guess. A good improvement though would be to list *all* uses of a blob. ssb-meme is giving us some uses that we don't show and that would be easy to change https://github.com/fraction/oasis/pull/425/files#diff-742340a98a2219160a3a98fd3a615335R1052 , it would take a bit more effort to show *all* uses of a blob but we could do that.

> Above the image grid, add some text "Please look at the original post to see if the author is OK with reposting their image"

Good idea, will (hopefully remember to) do this when I/someone adds the embed stuff in another PR. (right now in oasis it's pretty hard I think to copy an image from one post to another)

> When copy-pasting the blob address, also include an attribution markdown link that will end up getting pasted into the new post. The user might choose to remove that link before publishing, but it would set a cultural norm around attribution.

Maybe personal preference but I'd prefer if the client made use-finding easier in an automatic way, for instance like patchwork has `right-click image -> show references`.

> How does this interact with blobs in private messages?

Depends on how ssb-meme [works](https://github.com/ssbc/ssb-meme/blob/master/index.js). I'm not good at FlumeView stuff ([flumeview-search](https://github.com/flumedb/flumeview-search) in this case) but my bad guess is that they're not included. Do you have a worry?

package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@fraction/base16-css": "^1.1.0",
-    "@fraction/flotilla": "^4.0.1",
+    "@fraction/flotilla": "github:Powersource/flotilla#add-meme-2",
```
Powersource, line 25: todo: use semver again once this is merged

christianbundy: just published as 5.2.0 :boom:

cinnamon-bun:
> Due to the way blobs work it's not provable who uploaded the blob initially, and without some major causality math (using links between feeds to figure out who posted when) I don't think we could even make a very good guess. 

Ah, you're right!  And without that, the rest of my ideas make less sense.

Maybe we could... only show images that were used in more than one post?  That would at least reduce the personal images that show up.

It doesn't look like ssb-meme tracks that usage number right now though.  This could be a future enhancement.

---

> Maybe personal preference but I'd prefer if the client made use-finding easier in an automatic way, for instance like patchwork has right-click image -> show references.

That'd be nice!  We can't change the right-click menu but could do this:
* User sees an image they want
* Right click --> copy URL using native browser menu, get an url like `http://localhost:3000/blob/%26ByhS...`
* Compose a markdown message referencing the image
* When we're about to post it, detect the Oasis format blob URL and convert it into a plain old blob reference

We should probably do the same detection and replacement of other Oasis URL formats (message and user pages)

Powersource:
Should be ready for merging now!

sbillig:
This is awesome, @Powersource. Looks and works great.

christianbundy:
WOOP WOOP

PR #427, opened 2020-05-04
Upgrade Flotilla and SSB-CONN, remove auto-stage

christianbundy:
Problem: The auto-stage code was a horrible hack that may be causeing an
OOM bug if too many peers are staged too quickly.

Solution: Since SSB-CONN now supports a variable number of connection
hops, we can remove the auto-stage hack and use the default scheduler
instead of trying to staple on our own hacky scheduler fix.


src/ssb.js
```diff
@@ -198,11 +134,25 @@ module.exports = ({ offline }) => {
...
+  // Make a copy of `ssbConfig` to avoid mutating.
+  const customConfig = JSON.parse(JSON.stringify(ssbConfig));
+
+  // This is unnecessary when https://github.com/ssbc/ssb-config/pull/72 is merged
+  customConfig.connections.incoming.unix = [
+    { scope: "device", transform: "noauth" },
+  ];
+
+  // Only change the config if `--offline` is true.
+  if (offline === true) {
```
nickwynja, line 146: When I run `npm run dev -- --offline` I get the following error:

```
/Users/nickwynja/Projects/oasis/src/index.js:107
    throw err;
    ^

TypeError: Cannot set property 'autostart' of undefined
    at module.exports (/Users/nickwynja/Projects/oasis/src/ssb.js:147:33)
    at Object.<anonymous> (/Users/nickwynja/Projects/oasis/src/index.js:139:16)
```

christianbundy: On it!

christianbundy:
Woo! Thanks for merging this.

PR #436, opened 2020-05-12
Adds view specific titles

### LINKED ISSUE #55: <title> of page is always just "Oasis" -- maybe include thread title
**What's the problem you want to solved?**
I like to open lots of Oasis tabs, and also bookmark them for later.  But they all have the same page title, "Oasis".

For message or thread views, put the title of the message/thread in the <title>.

Extracting the message title isn't easy, but maybe:
1. If there's any markdown headlines, use the first one of those
2. Otherwise remove markdown image references in case the message starts with an image, then use the first line of text

nickwynja:

Closed #55 by adding view specific `<title>`. 

Uses `i18n` translation or non-language specific attributes like author name or snippet. Instead of implementing any new `ssb-markdown` functionality as mentioned in the issue, I just used the existing `postSnippet()` function that's used in subtopic previews to get a snippet of the message content.

I would welcome any feedback on implementation or by happy to make any desire changes.

### DIFF
--- a/src/views/i18n.js
+++ b/src/views/i18n.js
@@ -165,10 +165,12 @@ const i18n = {
       " commented on ",
       a({ href: parentUrl }, " thread"),
     ],
+    commentTitle: ({ authorName }) => [`Comment on @${authorName}'s message`],
     subtopicDescription: ({ parentUrl }) => [
       " created a subtopic from ",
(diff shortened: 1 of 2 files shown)


christianbundy:
Oh this is a great improvement, thank you!

PR #438, opened 2020-05-13
Adds visual separator between different threads

nickwynja:

When using a text browser (w3m in my case), it's hard to distinguish between different threads in the Summaries and Threads view. The indentation and CSS-based left border on comments and subtopics don't render in text-based browsers.

There was very little visual distinction when the comments on a thread ended and when the next thread begins:

Screen Shot 2020-05-13 at 2 30 29 PM

Introduce a `<hr>` after a thread. This is determined by when a message `isRoot` and has `aside`.

Screen Shot 2020-05-13 at 2 30 07 PM

The `text-browser` class is added to `display: none;` in CSS. In effect, this PR makes no change to graphical browsers and only displays the horizontal rule when browsers that don't use CSS. This seems to be a better approach than my last attempt (#364).

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -131,6 +131,13 @@ section header a {
   margin-left: var(--size--2);
 }
 
+/* For use with elements specific for
+ * rendering in a text browser and intended
+ * to be hidden in a graphical browser. */
+.text-browser {
+  display: none;
+}
+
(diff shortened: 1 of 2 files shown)


christianbundy:
This is so clever! What a great fix.

PR #448, opened 2020-05-23
Add simple pagination to user feeds.

black-puppydog:
This is simply based on sequence numbers and the `gt`/`lt` arguments of
`createUserStream`.
The use of `footer()` for the pagination links is quite hacky but I didn't see how the links were made to be left/right aligned so this was my way out of that :shrug:


black-puppydog:
Boom, got it to lint and all. :P
Still unsure about the `footer()` hack but it does work for me and it's been pretty handy already. 
I understand this is not as universal a pagination method as giving `gt` and `lt` as message keys, but for user feeds this works and will always be enough as long as ssb sticks with the "append only linked lists" data model.

christianbundy:
Hey, thanks for putting this together! Just tried it out locally and I can confirm It Works. :star2:

PR #449, opened 2020-05-24
Add rendering of blog messages.

black-puppydog:
`blog` type messages were not getting rendered besides the generic JSON dump.

This PR simply loads the blob during the `transform()` call.

I'm unsure though how this everything should be handled in public mode.
The blog message itself might be private, or it might reference a boxed blob... what then?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -421,27 +421,7 @@ router
   })
   .get("/blob/:blobId", async (ctx) => {
     const { blobId } = ctx.params;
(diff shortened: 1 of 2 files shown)


black-puppydog:
test message: `%H1Z9w2inv5YSaqD4UNwsfRy+vGfovkrK1kl6zhIQVz4=.sha256`

christianbundy:
Thanks for this patch!

PR #451, opened 2020-05-25
Allow GET params in /blob calls.

black-puppydog:
This allows blobs to load when the blob reference in the markdown has
`?contentType=image/jpeg` or similar hints attached.
It does so by simply removing the GET parameters from the URL though.
It would be a bit nicer to actually use those hints to set the right
HTTP response headers, but somehow koa keeps ignoring / overwriting (?)
the manually-set `Content-Type` headers, so this has to do for now.


black-puppydog:
Test message: `%VeiuOtgOpA7yzzCFvoli6H19NQ2DRx/LJ4lPssi1nLk=.sha256`

christianbundy:
And this one too!

PR #461, opened 2020-10-08
Add absolute post time as tooltip to timestamps.

black-puppydog:
Oh, so Christian's birthday was "1y ago"? But when *exactly* was that?
Hover the timestamp to find out! :)

Needing to know the exact time something was posted.

A tooltip over the UI element that shows the relative time. It's a classic pattern I guess?

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -334,7 +334,10 @@ const post = ({ msg, aside = false }) => {
   );
 
   const { name } = msg.value.meta.author;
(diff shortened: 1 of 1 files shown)


black-puppydog:
@christianbundy or @cblhg or whoever: this is passing now. just didn't have the prettier hook set up on this machine...

christianbundy:
Beautiful, thanks for adding this! :tada:

PR #462, opened 2020-10-19
Post preview with blobs and mentions

cryptix:
This PR contains the outcome from `%9NXKauuXCgJCHUaIXFGbpnXucAmvZ0QyQFJOAzURspc=.sha256`: @cblgh & @cryptix's sprint to add **[post previews](https://github.com/fraction/oasis/issues/67)** and **[blob uploads](https://github.com/fraction/oasis/issues/265)** to the post composer. Next to those two big features @cblgh also implemented a variety of visual tweaks, including a more concise thread rendering, and we found enough time to improve the [`@mention`](https://github.com/fraction/oasis/issues/47) suggestion feature.

As such we think it should:
* closes #47
* closes #67
* closes #256

### Post preview
Screenshot_2020-10-19 Publish - Oasis

This is the preview flow we implemented. If you mention someone, and the match is ambiguous we show all the probable matches as cards, with a prepared markdown link that can be copied into the text. This is less complex and than an initial idea we had (which made use of `<input type:radio>` for each name).

### Thread view
oasis-threads1

Posts now stick together much closer. We think it makes it clearer what is a conversation with replies compared to the latest or popular view where it's just a list of (unrelated) posts.

## Miscellaneous 
**re oasis-desktop:**
Apart from one [weird bug](https://github.com/fraction/flotilla/pull/3#issuecomment-711822213) in the database server, it all went quite swell.. but that one is a bit _frightning_ as cryptix couldn't figure out where it's comming from. The TL;dr for that one is: it randomly stopped showing names and avatar images for (some) people..

Our current bit-sized summary action plan is: make a release with just the new features (as non of that requires updating the database) and make a -beta release with most recent dependencies to test this further.

**A Blobby Caveat**
A caveat with the current blob upload is that, after a file has been attached, the preview needs to be rendered (by clicking the _Preview_ button) before anything is visible in the composer. This can be slightly confusing. The alternative would be to rely on the browser native file input, which is notoriously hard to style.

**other nice things**
* Blobs use the correct markdown link based on the mime-type (i.e. supports `video:...` and `audio:...`)
* Blob uploads are possible in all views that public posts can be composed (the publish, comment, and subtopic views) 
* Exif location data is stripped from blob uploads
* We planned to use [ssb-blob-files](https://github.com/ssbc/ssb-blob-files/issues/16) for the exif stripping but ran into problems with it. This would be a nice follow-up for private-blobs, tho.
[...]

PR #465, opened 2020-11-11
Fix tests and remove workaround

christianbundy:

Problem: Recently there was a PR [0] merged with a quickfix to avoid
some test failures, which is something I've been trying to avoid. While
`process.exit()` works fine, I'm worried that it means we don't
understand what's happening under the hood, *plus* I have the [maybe
unjustified?] worry that it might kill the process during a database
write or something dangerous. It looks like this particular test hang
was caused by both a stream and some number of intervals that were left
open.

Solution: Provide a way to close the stream and intervals in `index.js`
and ensure that we do that before closing the server.

[0]: https://github.com/fraction/oasis/pull/462

cc: @cryptix @cblgh @black-puppydog @cinnamon-bun

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -4,6 +4,7 @@ const debug = require("debug")("oasis");
 const { isRoot, isReply: isComment } = require("ssb-thread-schema");
 const lodash = require("lodash");
 const prettyMs = require("pretty-ms");
+const pullAbortable = require("pull-abortable");
 const pullParallelMap = require("pull-paramap");
 const pull = require("pull-stream");
 const pullSort = require("pull-sort");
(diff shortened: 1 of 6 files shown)


src/models.js
```diff
@@ -306,6 +258,70 @@ module.exports = ({ cooler, isPublic }) => {
         })) || "";
       return raw;
     },
+    // This needs to run in the background but also needs to be aborted
+    // in index.js when the server closes. There's also an interval that
+    // needs to be cleared. TODO: Ensure that there's never more than
+    // one interval running at a time.
+    _startNameWarmup() {
+      const abortable = pullAbortable();
+      let intervals = [];
```
black-puppydog, line 267: I think the only thing I don't understand about this patch is this change, to have a *list* of intervals...

christianbundy: Potential problem: I didn't see any code that ensures that we only have one interval running at a time. I _think_ that the 'sync' events happen every time we're up-to-date, which gives me the impression that we might have multiple intervals active at once? I don't know, didn't have time to test but also *really* didn't want to miss any of these things.

Happy to merge an improvement removing the list if we can be sure there's only one interval active at a time.

black-puppydog: okay so after looking at this a bit harder.... I think I see why you made this into a list of intervals. you're expecting `msg.sync` to be `true` more than once?
If that's the case, then that should have already been the case before, no? 

What I don't see in your code though is something that would ever remove an interval id from the list again. 
Shouldn't the code that's triggered by the interval remove it from the list again?

christianbundy: > If that's the case, then that should have already been the case before, no?

Yep, exactly. Previously this interval wasn't being cleared, which is one reason why the tests hung forever -- this clears the intervals, but since I'm not sure how many there are then I'm double-checking to make sure there can't be more than one.

(This could also be accomplished with an `if`.)

> What I don't see in your code though is something that would ever remove an interval id from the list again.

Like this, or do you mean something else?

https://github.com/fraction/oasis/pull/465/files#diff-a19725de73a858e12dc293aef7c29e8d9d107497643f2c65175ab0064d804665R321

black-puppydog: That clears the interval, yes. But the integer (?) referencing this interval will still be held in the `intervals` variable, no?

black-puppydog: (sorry if I'm wasting your time. I haven't used these before, not really, and I'm just going by what I find in MDN)

christianbundy: That's right! But this only happens when the Oasis closes so we don't need to worry about removing items from memory. If we were clearing items every few minutes or so, you're absolutely correct that this would cause a memory leak. I'd be happy to merge a PR improving this to do it a different way, I mostly just want to fix up the tests.

black-puppydog: No, just checking that I understand what's happening. Clearing up a few bytes just before exit seems... non-prioritary :P

christianbundy:
Self-merging.

Issue #466, opened 2020-11-11
Cannot post a mention-like text without it actually becoming a mention


black-puppydog:
## What's the problem you want solved?

This happened in `%zfgm5l/INLI9Nggn7hzniQjzr2mr2aC9korQ/UpARTQ=.sha256`:
I tried to post this text:

```terminal
daan@hermes ~/code/patchwork 🦆127» npm ci                                                                                                                                                         master|✚1 16:46
npm ERR! code 1
```

The mentions flow made it into this:

```terminal
daan[@hermes](@hoSvAN2VxIumAjQp1jY7F4ZBujATFuKOgBtYEIXJc6o=.ed25519) ~/code/patchwork 🦆127» npm ci                                                                                                                                                         master|✚1 16:46
npm ERR! code 1
```

There are IMO two issues with this, but I think they may be worth thinking about together first, and breaking them up later:

1. This was clearly a false-positive. Arguably, code fragments rarely need mentions.
2. I noticed this before posting, but the UI didn't actually allow me to fix it by hand. I removed the matched mention, but then "Preview" re-introduced it. So I was left with hitting "Publish" which then published the message with the wrong mention.

## Is there a solution you'd like to recommend?

1. Don't match mentions in code (inline or block)
2. Provide some way of previewing/publishing a hand-fixed version.

cblgh:
1 is very doable (one approach: refine regex to not match inline code, temporarily splice out code blocks from text & reintroduce after matching mentions)

2 i have no suggestions for / feels difficult

i probably won't get to this for a while, so anyone else is most certainly free to prod it 🖤

christianbundy:
Another take on 1:

1. Markdown -> HTML -> DOM (jsdom/etc)
2. Walk the DOM and make substitutions, avoid making substitutions inside `code` or `pre` (or `img src` or whatever)
3. DOM -> HTML  -> Markdown

christianbundy:
Another option: weirder syntax.

Example: `<@christian>`

Sure, this isn't a Good Solution in that it doesn't solve for `<pre>` or `<code>` elements, but it's also a hell of a lot less likely to happen randomly. What do y'all think? :man_shrugging:

christianbundy:
I'd really love to get this quickfixed before the next release, what do y'all think the easiest solution is here? Also, does this mean that email addresses get replaced?

christianbundy:
[...]

PR #469, opened 2020-11-21
Use stricter regex for @mention match

christianbundy:

Problem: You can't publish email addresses or user@host combos or
anything like that because the regex is super liberal in what it
matches.

Solution: Ensure that the first word in a message is a mention *or*
ensure that there's whitespace before a mention. This also adds a
regex101 link (in lieu of proper testing, for now) which makes this
easier to test and track over time.

---

See-also: https://github.com/fraction/oasis/issues/466

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -158,14 +158,16 @@ const preparePreview = async function (ctx) {
   // This matches for @string followed by a space or other punctuations like ! , or .
   // The idea here is to match a plain @name but not [@name](...)
   // also: re.exec has state => regex is consumed and thus needs to be re-instantiated for each call
-  const rex = /(?!\[)@([a-zA-Z0-9-]+)([\s.,!?)~]{1}|$)/g;
-  //                                  ^ sentence ^
-  //                                   delimiters
+  //
+  // Change this link when the regex changes: https://regex101.com/r/j5rzSv/2
+  const rex = /(^|\s)(?!\[)@([a-zA-Z0-9-]+)([\s.,!?)~]{1}|$)/g;
+  //                                        ^ sentence ^
+  //                                         delimiters
 
   // find @mentions using rex and use about.named() to get the info for them
   let m;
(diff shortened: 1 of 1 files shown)


christianbundy:
cc: @cblgh

cblgh:
cant access regex101 atm, but the additional strictness (start of line or space character) looks good!

black-puppydog:
Yes, this looks useful! Even if it's not 100% and we'd still want to do other things, I'd want this.

black-puppydog (approved):

christianbundy:
THANKS! Please merge whenever @black-puppydog.

black-puppydog:
image

waiting to do one more test right now :P

black-puppydog:
one day the good people at github will allow us to select "semi-linear" merging aka "rebase and merge with merge commit". `*`sigh`*`

christianbundy:
what's the problem with a regular `git merge`?

black-puppydog:
https://github.com/isaacs/github/issues/1017
I think this one explains it well.

git history diagrams

black-puppydog:
It's a bit of a soapbox of mine. I've been drilling engineers at work to pay attention to the commit logs, because they're so damn useful for regression debugging.

christianbundy:
> I've been drilling engineers at work to pay attention to the commit logs, because they're so damn useful for regression debugging.

I couldn't agree more! But this is usually why I *hate* rebase-centric workflows, because they mutate commits and obfuscate what the real tree was during a PR. For example, maybe this PR only works with the `package-lock.json` from yesterday, and one of the new PRs that were just merged actually broke this commit -- if we use *any* rebase we're likely to mutate the history so that it looks like this commit seems like the broken one (instead of showing that it was on a branch that was fine because it had another parent).

Would love to discuss this with you, because I also feel very strongly about dev workflows. Maybe we should find another place to discuss tho? :laughing:

PR #470, opened 2020-11-21
Revert "Removes husky to fix #346"

### LINKED ISSUE #346: Remove pre-commit hook and rely on CI to ensure code quality

Using pre-commit hooks can be a frustrating developer experience and get in the way of someone's normal development workflow. It can have an unexpected outcome when saving work-in-progress as well as interrupt things like `git rebase` with several unnecessary runs of the linter.

Use the existing CI infrastructure to lint on PRs so that code quality is maintained without affecting developer workflow. Alternatively, it could be recommended that developer uses a non-blocking in-editor lint tool.

I'm still reading through [C4](https://hintjens.gitbooks.io/social-architecture/content/chapter4.html) but it looks like it mentions style guides but isn't specific to _how_ they should be implemented or enforced.

`master`

christianbundy:

Problem: I keep forgetting to run tests and my commits aren't actually
passing in CI. I'd really like for these to be run automatically so that
I don't have to remember my pre-commit hooks.

Solution: Revert commit d9e829e2e8eddcc7c8edae2dd542cb69d86836c1, which
Nick Wynja added to remove Husky as a pre-commit hook. I hope that
anyone who doesn't want to run the pre-commit hook will run Git with the
`--no-verify` option, and anyone bummed at this change will open a
conversation to discuss in more details (or just revert *this* commit).

---

See-also: https://github.com/fraction/oasis/issues/346

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,0 +1,5 @@
+{
+  "hooks": {
+    "pre-commit": "npm test"
+  }
+}
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -24,6 +24,9 @@ Please run `npm test` before writing a commit, because if there are errors then
 maintainers won't be able to merge your patch. Please ask for help if `npm test`
 is giving you any trouble.
 
+**Note:** `npm run fix` is run automatically as a pre-commit hook. You always
+have the option to disable pre-commit hooks with `git commit --no-verify`.
+
 ## Frequently Failed Tests
 
 ### Unknown word
--- a/package-lock.json
+++ b/package-lock.json
@@ -399,7 +399,7 @@
     },
     "@types/debug": {
       "version": "4.1.5",
-      "resolved": false,
+      "resolved": "",
       "integrity": "sha512-Q1y515GcOdTHgagaVFhHnIFQ38ygs/kmxdNpvpou+raI9UO3YZcHDngBSYKQklcKlvA7iuQlmIKbzvmxcOE9CQ=="
     },
     "@types/events": {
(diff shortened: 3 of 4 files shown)


black-puppydog:
I'm of two minds here...
On the one hand, yes, that's annoying.
On the other hand... I remember making my first commits and being confused and frustrated with `git gui` suddenly freezing etc...

christianbundy:
Thanks for the feedback. I'm surprised that `git gui` froze, what's that about? Was it running the pre-commit hook but not giving you any info about that?

black-puppydog:
Yes exactly. It was just unexpected and git gui is pretty basic UI after all.

christianbundy:
What do you think the best way to solve that would be? The new common-good (not merged yet) is super fast, and only tests the files you've staged, which should help -- maybe a note in the docs about the pre-commit hook? It seems like every time someone makes their first contribution we have to go over the `npm run fix` thing, and it'd be awesome to avoid that if possible without screwing power users with `git gui`.

black-puppydog:
Yeah I think a good **ALL CAPS BOLD WARNING** in Contributing.md might be enough.

christianbundy:
What were the changes you force-pushed?

black-puppydog:
just a rebase. I find merging from master into PRs exceedingly annoying to look at in the commit log :D

christianbundy:
In the future could we talk about that first? I was surprised that you force-pushed to my repo, especially with commits that modify my commit history.

black-puppydog:
Yes, sorry about that. I had kinda gotten used to the workflow from work, but you're right it's not the same here.
Won't happen again :)

black-puppydog:
I think next time I'll either comment & wait, or just make a new PR and @ you. That seems more polite than pushing to your repo. :P

christianbundy:
No biggie, thanks! :heart:

PR #495, opened 2020-11-22
Increase dependabot open PR limit

christianbundy:
Problem: 5 is not a lot and I'd rather have the builds all done and
ready when I'm ready to merge them instead of having to do them in
chunks of 5.

Solution: Allow 16 open PRs at a time. We can change as needed.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -18,9 +18,7 @@ jobs:
         os: [macos-latest, windows-latest, ubuntu-latest]
 
     steps:
-      - name: Set git to not change EoL
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -18,9 +18,7 @@ jobs:
         os: [macos-latest, windows-latest, ubuntu-latest]
 
     steps:
-      - name: Set git to not change EoL
-        run: |
-          git config --global core.autocrlf false
+        run: git config --global core.autocrlf false
```
christianbundy, line 21: oops, smol change to remove the silly name for this command

PR #507, opened 2020-11-22
Upgrade @koa/router

christianbundy:
This also fixes our routes so that we're listening on `/foo` instead of
`/foo/` with an accidental trailing slash.


christianbundy:
I think maybe this branch was the one breaking the mentions?? This should be checked (by me or a reviewer) before merging.

PR #509, opened 2020-11-22
Merge flotilla source code into Oasis

christianbundy:
Problem: I thought it would be a good idea to keep our SSB stack in
another module but it just creates friction and adds another layer to
our stack that has to be independently upgraded and versioned.

Solution: Merge the @fraction/flotilla source code into Oasis so that we
don't have to maintain two separate projects.


christianbundy:
I'm looking forward to having dependabot run on this repo. :mechanical_arm:

black-puppydog (approved):

black-puppydog:
This works for me. :)

PR #524, opened 2020-11-23
Revert CSS changes from preview PR

christianbundy:
Problem: There were lots of CSS changes in the preview + blob PR and I
didn't take enough time to look at them. It looks like a few of them
made style problems (e.g. extra specing in the 'next' / 'previous' links
on the profile pages) and it made it harder (IMO) to determine that
links are links (since we've been using font weight to distinguish).

Solution: Revert those changes back to the previous defaults. If the
margin glitch was on purpose or anyone feels strongly about
distinguishing links using some other style (color?) then I'm happy to
merge those, but I'd like to fix the extra margin and have *some* way to
distinguish links in the mean time.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -144,6 +144,7 @@ section header a {
   text-decoration: none;
   margin-right: var(--size--2);
   margin-left: var(--size--2);
+  font-weight: bold;
 }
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Also self-merging since this has been open for a couple days.

PR #532, opened 2020-11-25
Common good 4.0.3

black-puppydog:

#497 wasn't passing tests because the version bump was actually incompatible.

## What solution are you ~recommending~ capable of?

Fixing the common-good call, and ignoring the dev type and husky and stylelint dependencies... :P

### DIFF
--- a/.depcheckrc
+++ b/.depcheckrc
@@ -1,0 +1,5 @@
+ignores: [
+  "@types/*",
+  "husky",
+  "stylelint-config-recommended"
+]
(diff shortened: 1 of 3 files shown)


black-puppydog:
Also I need to figure out how to modify the dependabot PRs. I could prbly have just added this commit to #497

christianbundy:
Nice, thanks! No worries about pushing to the dependabot PRs, I haven't been doing that either. Plus it means that only one of us can work on a given update at a time. :+1:

PR #540, opened 2020-11-29
Add rebuild button to settings

christianbundy:
Problem: The latest release seems to have some index problems, which is
frustrating, but might be able to be circumvented with a database
rebuild. The rebuild was something that Ahau funded and I'm excited to
be able to finally expose this to users without asking them to `rm -rf`
random files in their `~/.ssb` directory.

Solution: Replaced unused 'progress' bars (always at 100%) with rebuild
button.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -710,7 +710,6 @@ router
   .get("/settings", async (ctx) => {
     const theme = ctx.cookies.get("theme") || config.theme;
     const getMeta = async ({ theme }) => {
-      const status = await meta.status();
       const peers = await meta.connectedPeers();
       const peersWithNames = await Promise.all(
         peers.map(async ([key, value]) => {
@@ -720,7 +719,6 @@ router
       );
 
       return settingsView({
-        status,
         peers: peersWithNames,
         theme,
         themeNames,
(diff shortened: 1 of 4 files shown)


skyebend:
Thanks! I will test this out, but first, do you have an idea/guess of what 'rebuild database' will change about the database?  (so I'll have a sense of what to look if it causes problems for Patchwork or other clients)

christianbundy:
Sure! I don't know if you've already read the description (included in this PR), but there are a bunch of 'indexes' that SSB uses to keep track of commonly used information. For example, everyone's preferred names. It would be slow to have to do name queries every time we want someone's name, so we keep a file somewhere in `~/.ssb/` that keeps track of everyone's names. Unfortunately this index is a common source of bugs -- sometimes the index is corrupted or somehow broken, which means that queries fail (or give incorrect info). Usually the easiest fix is to just rebuild the index again from scratch, which *should* produce the exact same result, but if there are bugs with the original index then usually the rebuilt index will work better.

Happy to answer questions if you have them.

christianbundy:
Self-merging.

Issue #543, opened 2020-12-04
Width on desktops is too restrictive


Perelandra0x309:
## What's the problem you want solved?
On desktops the width of the used view is about 1/3 of the available screen width on my 22" monitor

## Is there a solution you'd like to recommend?
I was able to modify the style.css to remove the max-width restriction.  It would be nice to have an option in the settings to set a width for the main content section.  There are two locations where a max-width is defined- lines 325 and 595.

For example, I changed:
```
  body {
    width: 100%;
    max-width: var(--measure);
    margin: 0;
  }
```
to 
```
  body {
    width: 96%;
    margin: 0;
  }
```

black-puppydog:
Hi @Perelandra0x309,
I actually agree with you on this one, but the column width was quite deliberately chosen. See `%/dyC6QfS7BZ55s4ISqvPIZCtsMCpPTDid33q/6ky7n0=.sha256` for discussion, including yours truly. :stuck_out_tongue: 

However, I'd love to have some css styling available to people who like to skin/mod their oasis. Maybe we could have a mechanism to load from `~/.config/oasis/custom.css` ?

Perelandra0x309:
Thanks for that link, that's exactly the dilemma.  I really like Blaine's post near the end talking about line height and readability.  He also mentions the example of people preferring different mouse tracking speeds, which I think is a perfect analogy.  Not everyone likes the same line width either, so this should be an adjustable setting.

Over many years I have wondered why monitors are landscape, and not portrait like books.  It makes no sense to me, especially now that most things are done on the web. If I am reading a news website it would be so much more natural to have it in the same format as a book.  I've contemplated attempting to change over to a portrait monitor layout but have yet to follow through.

Alternately, I can think of a 3 pane layout that would work well with landscape monitors, and still keep the actual posts to 1/3 or 1/2 screen width.  Adding in a pane that displays one entry for each thread (title only, or a summary with participants) then on another pane the selected thread's posts would all be displayed.  Sort of like the 3 paned email layout.  I still haven't gotten used to Oasis yet with the way it organizes things.  There seem to be a lot of ways to organize the thread (latest, topics, summaries, threads) that I feel like I am missing posts if I only stick to one or two.  It seems like I have to cycle through all 4 to see everything.  Having one pane display all the threads of those you follow that have posts in would be easier to keep track of recent posts.  But this is a different request really, only related in that it keeps the smaller line length that some people prefer.

The real solutions I see to this issue are either:
[...]

PR #544, opened 2020-12-06
Add support for linking an optional custom-style.css file from the user's oasis configuration folder.

Perelandra0x309:
Adjusting stylesheet values to change things such as the main content width requires editing the stylesheets included in the oasis source code.

Allowing the user to modify stylesheet values such as content width using a stylesheet file saved in their oasis configuration folder.

### DIFF
--- a/src/views/index.js
+++ b/src/views/index.js
@@ -1,5 +1,9 @@
 "use strict";
 
+const path = require("path");
+const envPaths = require("env-paths");
+const fs = require("fs");
+
 const debug = require("debug")("oasis");
 const highlightJs = require("highlight.js");
 
@@ -81,6 +85,20 @@ const template = (titlePrefix, ...elements) => {
       )
     );
 
+  const customCSS = (filename) => {
+    const customStyleFile = path.join(
+      envPaths("oasis", { suffix: "" }).config,
+      filename
+    );
+    try {
+      if (fs.existsSync(customStyleFile)) {
+        return link({ rel: "stylesheet", href: filename });
+      }
+    } catch (error) {
+      return "";
+    }
+  };
+
   const nodes = html(
     { lang: "en" },
     head(
@@ -88,6 +106,7 @@ const template = (titlePrefix, ...elements) => {
       link({ rel: "stylesheet", href: "/theme.css" }),
       link({ rel: "stylesheet", href: "/assets/style.css" }),
       link({ rel: "stylesheet", href: "/assets/highlight.css" }),
+      customCSS("/custom-style.css"),
       link({ rel: "icon", type: "image/svg+xml", href: "/assets/favicon.svg" }),
       meta({ charset: "utf-8" }),
       meta({
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -25,3 +25,22 @@ Which value is given is decided like this:
 1. If an argument is given on the command-line, use that value.
 2. Otherwise, use the value from the configuration file if present.
 3. If neither command-line nor configuration file are given, use the built-in default value.
+
+# Custom Styles
+
+The stylesheet values may be overridden by adding custom values to a file found in a
+standard folder for configuration, depending on your operating system:
+
+- Linux: `$XDG_CONFIG_HOME/oasis/custom-style.css`.
+  Usually this is `/home/<your username>/.config/oasis/custom-style.css`
+- Windows `%APPDATA%\oasis\custom-style.css`.
+- Mac OS, `/Users/<your username>/Library/Preferences/oasis/custom-style.css`
+
+As an example the width used for the main body may be changed to a different
+fixed width or a dynamic width:
+
+```css
+:root {
+  --measure: 75%;
+}
+```
--- a/src/index.js
+++ b/src/index.js
@@ -36,6 +36,24 @@ if (config.debug) {
   process.env.DEBUG = "oasis,oasis:*";
 }
 
+const customStyleFile = path.join(
+  envPaths("oasis", { suffix: "" }).config,
+  "/custom-style.css"
+);
+let haveCustomStyle;
+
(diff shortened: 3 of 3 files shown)


Perelandra0x309:
To resolve https://github.com/fraction/oasis/issues/543

src/views/index.js
```diff
@@ -81,13 +85,23 @@ const template = (titlePrefix, ...elements) => {
       )
     );
 
+  const customLink = (filename) => {
+    const customStyleFile = path.join(envPaths("oasis", { suffix: "" }).config, filename);
+    try {
+      if (fs.existsSync(customStyleFile)) {
+        return link({ rel: "stylesheet", href: filename });
```
black-puppydog, line 92: am I reading this would generate a link like this?

```html
<link rel="stylesheet" type="text/css" href="/home/daan/.config/oasis/custom.css">
```

If so, I don't think we can do it like that. Oasis is also run on public instances (most notably probably oasis-demo.fraction.io ) and there this would of course not work, but also try to load a local file into a public website.

I'm pretty sure modern browsers wouldn't allow us to do that anyhow, but it still seems iffy...

black-puppydog: I'd proposed to add a route to koa that simply fetches `custom.css`, returning an empty document if it doesn't exist.
Kind of like here:

https://github.com/fraction/oasis/blob/master/src/index.js#L617-L627

Specifically, the cache-control can probably be used here too.

Perelandra0x309: No the link will reference like:
<link rel="stylesheet" href="/custom-style.css">

And the server will "map" (sorry don't know the correct term) /custom-style.css to /home/daan/.config/oasis/custom-style.css because I added a mapping in src/index.js similar to the way /theme.css is mapped to a file in the @fraction/base16-css module.

/src/index.js line 570 in the commit, where customStyleFile is a const path to the user's file like /home/daan/.config/oasis/custom-style.css:
```js
  .get("/custom-style.css", (ctx) => {
    ctx.type = "text/css";
    ctx.body = requireStyle(customStyleFile);
  })
```

Regarding your reference to loading a blob, I am not entirely sure what is happening with that code, but two things to ask.  Is it returning a 404 on file not found or empty?  That was what I was trying to avoid by checking if the file exists before adding a stylesheet link, so that you don't get a 404 error in the console if the custom-style.css file does not exist.  Not having that file is perfectly acceptable and that error would be confusing and also waste resources for the server trying to get a file which does not need to exist.

Second, am I reading the blob code that is caches the file?  I don't think caching would be a good idea for a css, if I am editing the css file to make tweaks I don't want it to be cached, but see changes I've made immediately on the next refresh.

Sorry this is my first time with node.js code, looks like there are two places that failed checks but do run fine in testing.  I can work on getting those cleaned up.

black-puppydog: Oh sorry, indeed I skipped those lines with the handler, that looks good then. :)
It will require restarting oasis the first time that the css file is created (to set `haveCustomFile`) but that's fine by me. 

for the cleaning up, I think simply calling `fs.readFileSync(customStyleFile, "utf8");` (without the assignment) should do.

src/views/index.js
```diff
@@ -81,13 +85,23 @@ const template = (titlePrefix, ...elements) => {
       )
     );
 
+  const customLink = (filename) => {
```
johanbove, line 88: Wouldn't it be better if `customLink()` would keep the same parameter signature as `link()`, so in that sense it should accept an object like this example as an argument: `customLink({ rel:"stylesheet", href: "/custom-style.css"})`. Otherwise I would rename the function to `customCSS()`.

Perelandra0x309: Yes that makes sense since a link can be used for other objects.  This is a specific case for a custom CSS file, I don't see this being used for other links links icons.

black-puppydog:
@Perelandra0x309 the commit log for this is a bit of a mess right now. From what I can see, you created a new commit, then a merge-commit to join the two attempts?
I cleaned this up locally by...

* creating a new branch based on `master`
* cherry-picking the later commit (the one with your modifications to shut up the linter)
* checking out the `add-custom-style` branch, and hard-resetting it to the now cherry-picked commit

I could force-push this but it is considered (rightfully, IMHO) a bit impolite to force-push into other people's repos, so I'll let you do this. Apart from that, this looks good to me, and I just tried it and love it!
Thank you for making this. :)

black-puppydog:
BTW the failing test is just `prettier` complaining about formatting. `prettier -w .` will fix that.

Perelandra0x309:
> @Perelandra0x309 the commit log for this is a bit of a mess right now. From what I can see, you created a new commit, then a merge-commit to join the two attempts?

Yeah I tried to do an amended commit but then git complained that the local branch was behind remote.  Does just making a PR modify the remote branch?  I tried to correct it and ended up with a mess.  Git really irks me sometimes.

OK I made an attempt at a fix, creating a new branch with the final changes, renaming it then doing a force push.  It looks like the PR sees just one commit now.

It looks like there still are some formatting issues.  I haven't found a set of instructions that get prettier working on my machine yet.  By eye the code looks like it conforms to the rest but there are no indications in the build check exactly what it doesn't like.

Perelandra0x309:
Oh damn I forgot to include the documentation.

Perelandra0x309:
OK I think this PR is ready now if the formatting warnings are not too egregious.  Thanks for putting up with a first time node.js hacker (I'm not a true developer).

christianbundy:
It looks like `npm test` is failing, would you mind running `npm run fix` to pass the formatting tests please?

black-puppydog (approved):

black-puppydog:
Hooray, this works for me and all the tests/linting/formatting issues are gone, too! Thank you for this @Perelandra0x309!

PR #584, opened 2021-01-26
Issue 126: Show recipients of private messages

lindskog:
In a private message you can't see who the recipients are, so you can't know the audience you're talking to.

Shows recipient thumbnails in the top right corner of each message (see image)

image

### DIFF
--- a/src/models.js
+++ b/src/models.js
@@ -693,8 +693,6 @@ module.exports = ({ cooler, isPublic }) => {
   };
 
   const getUserInfo = async (feedId) => {
-    const id = feedId;
-
     const pendingName = models.about.name(feedId);
     const pendingAvatarMsg = models.about.image(feedId);
 
@@ -708,8 +706,16 @@ module.exports = ({ cooler, isPublic }) => {
 
     const avatarUrl = `/image/64/${encodeURIComponent(avatarId)}`;
 
-    return { name, id, avatarId, avatarUrl };
+    return { name, feedId, avatarId, avatarUrl };
   };
+
+  function getRecipientFeedId(recipient) {
+    if (typeof recipient === "string") {
+      return recipient;
+    } else {
+      return recipient.link;
+    }
+  }
 
   const transform = (ssb, messages, myFeedId) =>
     Promise.all(
@@ -853,8 +859,8 @@ module.exports = ({ cooler, isPublic }) => {
 
         if (isPrivate(msg)) {
           msg.value.meta.recpsInfo = await Promise.all(
-            msg.value.content.recps.map((feedId) => {
-              return getUserInfo(feedId);
+            msg.value.content.recps.map((recipient) => {
+              return getUserInfo(getRecipientFeedId(recipient));
             })
           );
         }
--- a/src/models.js
+++ b/src/models.js
@@ -691,6 +691,26 @@ module.exports = ({ cooler, isPublic }) => {
       }
     });
   };
+
+  const getUserInfo = async (feedId) => {
+    const id = feedId;
+
+    const pendingName = models.about.name(feedId);
+    const pendingAvatarMsg = models.about.image(feedId);
+
+    const pending = [pendingName, pendingAvatarMsg];
+    const [name, avatarMsg] = await Promise.all(pending);
+
+    const avatarId =
+      avatarMsg != null && typeof avatarMsg.link === "string"
+        ? avatarMsg.link || nullImage
+        : avatarMsg || nullImage;
+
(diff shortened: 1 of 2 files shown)


christianbundy:
Thanks for taking on this work! I'm getting an error when I use this commit and try to visit `/inbox` though:

```
/home/christianbundy/src/fraction/oasis/src/index.js:132
    throw err;
    ^

Error: can only encode arrays
    at encode (/home/christianbundy/src/fraction/oasis/node_modules/charwise/codec/object.js:27:44)
    at encodeItem (/home/christianbundy/src/fraction/oasis/node_modules/charwise/codec/object.js:41:20)
    at Object.encode (/home/christianbundy/src/fraction/oasis/node_modules/charwise/codec/object.js:33:24)
    at Object.exports.encode (/home/christianbundy/src/fraction/oasis/node_modules/charwise/index.js:39:28)
    at Codec.encodeKey (/home/christianbundy/src/fraction/oasis/node_modules/encoding-down/node_modules/level-codec/index.js:32:45)
    at /home/christianbundy/src/fraction/oasis/node_modules/encoding-down/node_modules/level-codec/index.js:72:14
    at Array.forEach (<anonymous>)
    at Codec.encodeLtgt (/home/christianbundy/src/fraction/oasis/node_modules/encoding-down/node_modules/level-codec/index.js:70:21)
    at new Iterator (/home/christianbundy/src/fraction/oasis/node_modules/encoding-down/index.js:115:26)
    at DB._iterator (/home/christianbundy/src/fraction/oasis/node_modules/encoding-down/index.js:102:10)
```

This error only happens when I go to `/inbox`, but I'm not sure why.

Powersource:
@christianbundy I think that might be an unrelated error, see `%pWLkL41WTXzesxNVqUGriSCe9jI+SwdJAJv33X4eFsc=.sha256` . Good news is that cel says "this issue is fixed in latest ssb-server v15", "crash when handling certain messages recently published", bad news is that we don't use ssb-server directly so we'd have to do some more digging.

christianbundy:
:beetle:  Found the bug, it's in this pull request.

```diff
diff --git a/src/models.js b/src/models.js
index 92a0750d..c73d4c27 100644
--- a/src/models.js
+++ b/src/models.js
@@ -693,6 +693,7 @@ module.exports = ({ cooler, isPublic }) => {
   };
 
   const getUserInfo = async (feedId) => {
+    console.log({ feedId })
     const id = feedId;
 
     const pendingName = models.about.name(feedId);
```

# Pass

```js
{ feedId: '@8EDDP8C8+fAYjYKLATHzjns1e21J4YftUT+RrDu0XRk=.ed25519' }
```

# Fail

```js
{
  feedId: {
    link: '@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519',
    name: 'Christian Bundy'
  }
}
```

Previously we were reading `msg.value.author`, which is always a string, but `recps` is the wild west and we need to practice some defensive driving to make sure we're passing a string rather than an `{ link, name }` object. I think there's somewhere else in the Oasis code that handles this, maybe we could reuse that?

christianbundy:
@Powersource I don't think we use SSB-Links, so I don't think that bug is possible in Oasis. It *is*, however, a similar bug: https://github.com/ssbc/ssb-links/commit/d4916ed1c57c5ab055eb011bbc3a95a75a81eb97

christianbundy:
Example solution:

```javascript
const variants = [
  "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519",
  {
    link: "@+oaWWDs8g73EZFUMfW37R/ULtFEjwKN/DczvdYihjbU=.ed25519",
    name: "Christian Bundy",
  },
];

function getRecipientFeedId(recipient) {
  if (typeof recipient === "string") {
    return recipient;
  } else {
    return recipient.link;
  }
}

variants.map(getRecipientFeedId).every((x) => typeof x === "string"); // => true

lindskog:
Oh wow thanks for noticing that bug!! I will look into it right away, I just assumed that feedId was always a String

lindskog:
This is the issue for reference: https://github.com/fraction/oasis/issues/126

lindskog:
@christianbundy I've committed a suggestion for a fix to the bug! But I wasn't able to reproduce the bug myself (I had to hardcode the feedId to get the error that you mention), so please double check that the bug is actually solved (and that my solution for it is ok) :)

christianbundy:
This works great, thanks fr putting this patch together!

Issue #399, opened 2020-04-13
Can see posts from blocked people in threads


Powersource:
## What's the problem you want solved?

Links and screenshots from my current installation (latest git):

http://localhost:3000/thread/%256ligxbr4DgHxlTjCCQNQaH%2FF3xOSmv7AQsBiLymfEcY%3D.sha256#%256ligxbr4DgHxlTjCCQNQaH%2FF3xOSmv7AQsBiLymfEcY%3D.sha256

![image](https://user-images.githubusercontent.com/1427063/79125072-d085f980-7d9d-11ea-86b6-864390477a8a.png)

http://localhost:3000/author/%409BbJwPDjcyIqrOUPNn0nJZBduWdIrpMk3Cjz5MP361s%3D.ed25519

![image](https://user-images.githubusercontent.com/1427063/79125138-e8f61400-7d9d-11ea-80c3-2b57a2989df2.png)

These two pages render with the same state, i.e. I haven't clicked anything other than his name to get to his profile.

## Is there a solution you'd like to recommend?

Fix the bug somehow.

christianbundy:
My guess is the thread view (not the 'Threads' view) needs to be run through `socialFilter()`.

cbleslie:
@christianbundy Show Stopper? Hard to call, is this something a migrating user could ignore? Depends on the development lift maybe?

Powersource:
I think this is a pretty basic issue and I wouldn't be happy recommending oasis to friends without it

christianbundy:
Yep. It's in the barebones milestone and labeled "bug" and "safety". What are you recommending?

Powersource:
Yeah was replying to Cameron. I'm still recommending Patchwork to people :/

lindskog:
Just want to let everyone know that I just started working on this issue :)

PR #585, opened 2021-01-27
Add social filter to threads

lindskog:
Can see posts from blocked people in threads (Issue https://github.com/fraction/oasis/issues/399)

Add a social filter to threads


christianbundy:
Beautiful! Nice work.

PR #603, opened 2021-02-13
updating code for likes page bug

### LINKED ISSUE #110: The likes page shows stuff that's not liked anymore

SomaRasu:
Fixes #110 
Updated pendingVoters and added filter to getting "likes"


Powersource:
Seemed like this was working when we were testing locally :)

Powersource:
This works well except for https://github.com/fraction/oasis/issues/614 which also happens on master so I'll merge it :)

Issue #537, opened 2020-11-27
message Preview seems to submit the contents of the preview, even if additional edits have been made


skyebend:
## What's the problem you want solved?
When editing post in /publish, I use Preview (thanks for this amazing feature!) and notice typos.  I corrected my typos, but then clicked Publish without clicking Preview again, the version that was published did not include my new corrections.

## Is there a solution you'd like to recommend?
If Publish is pressed and the state of the text editor field does not match the previous preview state, should generate a warning for user like "Please preview your message before publishing" ?

lindskog:
Working on this issue :)

PR #611, opened 2021-02-17
Separate edit and preview view for posts and comments

lindskog:
Message preview seems to submit the contents of the preview, even if additional edits have been made (issue https://github.com/fraction/oasis/issues/537)

Instead of the recommended solution in the issue I discussed this with some people and suggest that edit and preview of posts and comments are separated into two different views, to create a more natural and less confusing user/data flow.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -131,6 +131,12 @@ button,
   color: var(--fg);
 }
 
+.edit-button {
+  float: right;
+  margin: 0;
+  margin-top: -1.9rem; /* This is not the best way to solve this but I couldn't think of anyother way atm, please let me know if anyone have any suggestions :) */
+}
+
 #blob {
   visibility: hidden;
   height: 0;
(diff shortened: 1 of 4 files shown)


Powersource:
Thanks for the PR!

Tried it out and seems to work well.

But the user flow when mentioning someone gets a bit weird since the people you might mention are showed to you on the preview screen and if you want to mention them you have to copy the mention and then click Edit to go back and use the mention. But maybe this is fine for now, overall an improvement and we should merge for now, what do you think @christianbundy ? In another PR we could maybe add another button next to the Preview button, that says "Get mentions" or something, that keeps you on the edit screen but lists the possible mentions for you?

image

christianbundy:
Thanks for working on this issue. I agree that this is a step backward for mentions, is it possible to implement this in a way that doesn't break that feature? If you'd like, I'm happy to discuss on GitHub so that we can come up with a solution that'll work. I didn't see any conversation in #537.

Issue #622, opened 2021-03-01
Oasis shows replies and comments from blocked people


nickwynja:
## What's the problem you want solved?

The comments on a thread show content from people marked as blocked.

## Is there a solution you'd like to recommend?

I've taken a swing at fixing this but I'm all turned around. I've attempted to `filter` the thread based on the result of `models.friend..getRelationship(msg.value.author)` but am struggling with JS here.

From [models.js](https://github.com/fraction/oasis/blob/26b1e5c9b2dd21e572e53d3bc7d14d46b2b1e795/src/models.js#L1251-L1260):

```js
pullParallelMap(async (message, cb) => {
  // Retrieve a preview of this post's comments / thread
  const thread = await post.fromThread(message.key);
  thread.filter((msg) => {
    models.friend
      .getRelationship(msg.value.author)
      .then((rel) => {
          debug(rel);
          if (rel.blocking == true) {
            debug(msg)
          }
      })
  })
  lodash.set(
    message,
    "value.meta.thread",
    await transform(ssb, thread, myFeedId)
  );
  cb(null, message);
}),
```

Obviously, I'm just debugging here instead of doing anything practical but every which way I tried to apply this filter didn't work. Happy to keep chipping away on something and open a PR if anyone is able to point me in the right direction.

nickwynja:
I have a working version by taking a different approach and would some feedback on it. In `transform()`, I've added a value of `value.meta.blocking` to each message based on `getRelationship()`. Then I can use a simple filter:

```js
const t = thread.filter(msg => !msg.value.meta.blocking);
```

The upside of this is that `msg.value.meta.blocking` could be used in various places like in the UI. Any downsides to this approach?

Powersource:
This should have been fixed by https://github.com/fraction/oasis/pull/585 , are you sure you're running latest master?

nickwynja:
That only fixes not displaying root posts from blocked people. A problem that pops up all throughout Oasis (which I believe @georgeowell has mentioned in the past) is that the `socialFilter` only applies to the first level of items in the list. Comments/replies/children don't seem to be processed by the `socialFilter`.

I'm playing a bit more with the `blocking` attributes in the UI. It might help as a blanket case until we're able to get the queries to filter out consistently. It's a bit hacky but is a start.

![image](https://user-images.githubusercontent.com/1012017/109567817-f7cbdf00-7ab3-11eb-861b-8257398c5dbc.png)

nickwynja:
[...]

PR #624, opened 2021-03-02
Hides content of messages from blocked people

nickwynja:

#622 

This introduces a basic implementation to not display the content of
messages. It still shows that a message exists but shows localized
copy indicating why the content was hidden.

We should consider additional filters or queries in different views to actually
remove these posts but I'll leave that as up for discussion since I can see
value in showing that a blocked comment exists in a conversation.

### DIFF
--- a/src/assets/style.css
+++ b/src/assets/style.css
@@ -667,6 +667,10 @@ section.post-preview > section > footer {
   display: none;
 }
 
+section.post.blocked {
+  font-style: italic;
+}
+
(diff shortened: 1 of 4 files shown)


SomaRasu:
This PR looks good to me: I tested the code locally by blocking a user in a thread. Their message isn't displayed and a placeholder div saying "This message hides content from a blocked user." is shown instead.

christianbundy:
Beautiful, nice work @nickwynja!

Issue #217, opened 2020-02-16
Implement some kind of view for channels


georgeowell:
## What's the problem you want solved?
Patchwork has an emphasis on both recently active channels and subscribed channels. In Oasis, you can currently view channels (or hashtags) only by clicking through from a thread or searching for them.

Channels are especially good for new users to find others who have similar interests.

## Is there a solution you'd like to recommend?
That we consider how we can implement some views for channels. I'm open to ideas on this.

christianbundy:
+1 for supporting a list of hashtags, although I'm conflicted on 'channels' where you can only post to one topic at a time. This isn't something that Oasis currently supports, but maybe we should? I think I like the simplicity of using hashtags in the message rather than setting a channel in the message metadata, especially when it comes to weird stuff like 'what if someone changes the channel mid-thread?'.

georgeowell:
Yeh I think I am getting confused about the difference between channels and hashtags and tags. I assume everyone else is confused about this also :sweat_smile:

jedahan:
Same! I've also seen it written as 'topics' somewhere.

georgeowell:
I'm quite unsure how is best to present this list of hashtags to the user. So many options....

cinnamon-bun:
Some historical context about these SSB features in case anyone is wondering:

* `Hashtags` are autodetected when written in the message body.  Works like Twitter.
* `Channels` are assigned in message metadata by the author.  One per message.  I think this was intended to work like Reddit, where each channel would be viewed on its own page.  It would create warrens that tuck each channel's messages away from the main view, creating a new social context, a smaller space.  This also makes it easier to say "hey devs, please keep your programmer talk in the SSB channel so we can ignore it".  This never quite worked out because Patchwork shows channel'd messages in the main feed, and ended up conflating them with hashtags.
* `Tags` are attached to a post after it's written, by anyone (not just the author).  Possible use-cases are tagging NSFW things, organizing posts so you can find them later, bookmarking?  They have technical issues -- there can be duplicate tags with the same name but a different internal ID, which leads to a confusing user experience.
* `Topics` are not an official thing.  Christian uses this to mean "root-level posts" I think?

You can also `subscribe` to channels (and hashtags), and this brings those posts into your feed from people you're not following.  So they are a way to find new people beyond your usual social horizon.

cinnamon-bun:
My feeling is:
* Hashtags are familiar to users and widely used on SSB already, so let's support them.  Pretty much make them work like Twitter.
[...]

PR #626, opened 2021-03-03
Adds subscribed channels to Topics

nickwynja:

Subscribed channels make it easy to stay connected to topics you're interested in. Oasis doesn't display a list of topics anywhere so, as mentioned in #217, you need to access through search and then click on a hashtag.

Any subscribed channels now appear at the top of Topics. This gives you
easy access to view subscribed channels using the existing hashtags
view. Much more can be done to fully support features discussed in #217 but this is a start.

image

For anyone reviewing, you can subscribe and unsubscribe to channels with these CLI commands:

```
ssb publish --type channel --channel oasis --subscribed
```
and:

```
ssb publish --type channel --channel oasis --no-subscribed
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -471,7 +471,12 @@ router
   })
   .get("/public/latest/topics", async (ctx) => {
     const messages = await post.latestTopics();
-    ctx.body = await topicsView({ messages });
+    const channels = await post.channels();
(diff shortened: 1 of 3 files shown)


christianbundy:
Sure! This looks fun. Thanks for putting it together.

PR #647, opened 2021-03-30
docs: Fix Docker port mapping

asymmetric:

The oasis wouldn't be reachable at `localhost:3000`.

Use 127.0.0.1 for the host side of the mapping (localhost), rather than the incorrect 127.0.0.0.


christianbundy:
Nice! Thanks for this fix.

PR #784, opened 2022-03-19
Replacing `get` for `graph` in ssb-friends usage

nsantini:
Issue #718 reported the issue we got after bumping `ssb-friends` to 5.0.0.

This solution replaces the use of the legacy `get` for the new version `graph`. It creates a Promise with the resolved graph and filters it by just the current `id`.
It also updates what is considered "following" and "blocking" based on the new version specs


nsantini:
@christianbundy @cinnamon-bun Are you guys still actively maintaining Oasis?
I'm keen to try to solve more issues when I find time...

nickwynja:
@nsantini I should be able to take a look at this today. There's not a lot of activity here but a few people are able to merge PRs when others contribute.

nickwynja:
@nsantini I'm having trouble tracking it down exactly but it appears to me so far that no data is syncing when I have this PR checked out. I appear to be connected to peers but no new data is streaming in. Can you double check that you're not seeing that same behavior?

nsantini:
One thing I should improve here: add backwards compatibility. If you are running Oasis on top of say Patchwork, we might encounter the opposite bug, not having access to the new API

I have not encountered the issue you mention @nickwynja

nickwynja:
@nsantini What version of node are you on while running this branch?

nsantini:
Currently v14.16.0, but I think I was running v12 when I did that work

nickwynja:
I can't figure this out. My guess is you're following some pub or room that is replicating while I don't think I have any connections like that.

nickwynja:
I discovered #785 in the closed PRs. With the addition of the other bumped versions, things replicate fine. So I've added those changes and will merge.

